PR #3, opened 2021-10-21
add pallet teerex and CI

haerdib:
copy paste from master branch pallet teerex: https://github.com/integritee-network/pallet-teerex/blob/master/src/lib.rs

closes https://github.com/integritee-network/pallet-teerex/issues/54


brenzi (approved):
checked tests manually. all seems to be there
LGTM

PR #12, opened 2021-11-08
Feature/parentchain pallet

gaudenzkessler:


parentchain/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+
+	/// Hash of the previous block.
+	#[pallet::storage]
+	#[pallet::getter(fn parent_hash)]
+	pub(super) type ParentHash<T: Config> = StorageValue<_, T::Hash, ValueQuery>;
+
+	/// Map of block numbers to block hashes.
+	#[pallet::storage]
+	#[pallet::getter(fn block_hash)]
+	pub(super) type BlockHash<T: Config> = StorageValue<_, T::Hash, ValueQuery>;
```
haerdib, line 34: The comment and the storage do not match. What you have here is not a Map from BlockNumber -> BlockHash but just the storage of the last BlockHash. For a Map you need to set the BlockNumber as a key.

Example of a storage map: https://github.com/integritee-network/pallets/blob/master/teeracle/src/lib.rs#L55-L57

parentchain/Cargo.toml
```diff
@@ -0,0 +1,100 @@
...
+    'frame-system/std',
+    'sp-core/std',
+    'sp-io/std',
+    'sp-runtime/std',
+    'sp-std/std',
+]
+
+runtime-benchmarks = [
+    'frame-benchmarking',
+    "hex-literal"
```
haerdib, line 99: We have not a definite way of using ` '` or `"`, but lets stay consistent within the same file at least. I've looked around a little bit, and I think the prefered way is to use `"`

parentchain/src/tests.rs
```diff
@@ -0,0 +1,47 @@
...
+			H256::from_low_u64_be(1337)
+		));
+		assert_eq!(Parentchain::block_number(), 3);
+		assert_eq!(Parentchain::parent_hash(), H256::from_low_u64_be(42));
+		assert_eq!(Parentchain::block_hash(), H256::from_low_u64_be(1337));
+	})
+}
+
+#[test]
+fn verify_requires_root_works() {
```
haerdib, line 39: According to @clangenb the preferred way would be to use `xxx_errs` instead of `verify_xxx` in case of error testing

parentchain/src/tests.rs
```diff
@@ -0,0 +1,47 @@
...
+use sp_keyring::AccountKeyring;
+use sp_runtime::DispatchError::BadOrigin;
+
+#[test]
+fn verify_storage_works() {
+	new_test_ext().execute_with(|| {
+		assert_ok!(Parentchain::insert_block(
+			Origin::root(),
+			3,
+			H256::from_low_u64_be(42),
```
haerdib, line 29: This might be a personal preference, but I think in case of using the same variable twice in testing it makes sense to allocate it. It will make things easier to adapt in case the tests need to be changed. For example parent_hash needs to be removed from insert block, it'd be clear which one is the parent_hash in the test fn and the compiler will throw errors in case something is forgotten to be deleted.. in this case the `assert_eq!` from below.

haerdib (changes requested):

parentchain/Cargo.toml
clangenb: We actually have consensus to use `"` but it might not be consistent yet.

parentchain/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+	/// Configuration trait.
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		type WeightInfo: WeightInfo;
+	}
+
+	/// The current block number being processed. Set by `execute_block`.
+	#[pallet::storage]
+	#[pallet::getter(fn block_number)]
+	pub(super) type Number<T: Config> = StorageValue<_, T::BlockNumber, ValueQuery>;
```
clangenb, line 24: I would like if  the storage type and the getter always have the same name. Here, it'd be: `BlockNumber`.

parentchain/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+	#[pallet::getter(fn block_hash)]
+	pub(super) type BlockHash<T: Config> = StorageValue<_, T::Hash, ValueQuery>;
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight(<T as Config>::WeightInfo::insert_block())]
+		pub fn insert_block(
```
clangenb, line 42: As we don't 'insert` something into a map, but rather set a value, I'd go here for: `set_block`.

Also, I wonder if we should change the function signature to take a `Block` or `Header` as an argument (not sure here). I believe I could make sense as we have a more robust interface (not as sensible to code changes).

clangenb (commented):
Looks in general very good. Only minor comments and one thing that we might want to get consensus first.

parentchain/src/lib.rs
gaudenzkessler: I didn't call it set_block because I got a warning. 

"methods called `set_*` usually take self by mutable reference; consider choosing a less ambiguous name"

Can we find a name that fulfills your request and solves the warning🥴?

clangenb: I only get it in the IDE, do you also get it while compiling?

gaudenzkessler: No you are right. Only in the IDE. -> I'll change it. 

About your second question: Bigna said she also thought about it, but did not come to a conclusion yet 🤣
For me it is difficult to discuss about that. I do not have a strong oppinion

haerdib: I was also thinking about this, so lets change it. Sorry @gaudenzkessler, I hope that's okay for you.

 Our current `update_states` function in the worker (https://github.com/integritee-network/worker/blob/17c7624f00ad4cf4b0739f9703d43217bb61d188/enclave-runtime/src/lib.rs#L737-L740), where I'd suggest to place the parentchain_pallet call, takes the header as an argument. If you, @clangenb, also don't see a reason to pass the whole block, we should probably just use the header.

parentchain/src/lib.rs
haerdib: Are you sure about that @clangenb ? The frame_system pallet also uses `Number` (https://github.com/paritytech/substrate/blob/master/frame/system/src/lib.rs#L607-L610). As this pallet is a simplified system pallet, should we not use the same naming conventions?

parentchain/Cargo.toml
haerdib: Oho, good to know about that. Mabye we should write a StyleGuide @gaudenzkessler 😆

parentchain/src/lib.rs
clangenb: Ok, I agree that we should not change conventions here. I am just very surprised by the conventions that substrate uses in that case. :) fine with we, let's leave it.

parentchain/src/lib.rs
clangenb: Yes, it should be the header.

parentchain/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+sp-keyring = { git = "https://github.com/paritytech/substrate.git", branch = "master", version = "4.0.0-dev" }
+pallet-balances = { git = "https://github.com/paritytech/substrate.git", branch = "master", version = "4.0.0-dev" }
+
+[dev-dependencies]
+hex-literal = "*"
+env_logger = "0.7.1"
+log = "*"
+
+[features]
+default = ['std']
```
haerdib, line 34:
```suggestion
default = ["std"]
```

parentchain/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+std = [
+    "serde",
+    "codec/std",
+    "frame-support/std",
+    "frame-system/std",
+    "sp-core/std",
+    "sp-io/std",
+    "sp-runtime/std",
+    "sp-std/std",
```
haerdib, line 43: Could you also add `"log/std"` here?

parentchain/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+frame-support = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master", version = "4.0.0-dev" }
+frame-system = { default-features = false, package = 'frame-system', git = "https://github.com/paritytech/substrate.git", branch = "master", version = "4.0.0-dev" }
+sp-core = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master", version = "4.0.0-dev" }
+sp-io = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master", version = "4.0.0-dev" }
+sp-runtime = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master", version = "4.0.0-dev" }
+sp-std = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master", version = "4.0.0-dev" }
+frame-benchmarking = { default-features = false, optional = true, git = "https://github.com/paritytech/substrate.git", branch = "master", version = "4.0.0-dev" }
+sp-keyring = { git = "https://github.com/paritytech/substrate.git", branch = "master", version = "4.0.0-dev" }
```
haerdib, line 25: Now it's muuuuch nicer to look at. And easier to review, which explains why I only just now noticed: You're importing std dependent libraries here, which will not work in the sgx-runtime. It needs to be std free.
So either make it optional and import it only in case the std feature is activated or import it with
`default-features = false`.

For keyring, no no_std import is possible : https://github.com/paritytech/substrate/blob/master/primitives/keyring/Cargo.toml

For pallet-balances there is: https://github.com/paritytech/substrate/blob/master/frame/balances/Cargo.toml#L30-L40

haerdib: now the question is: What do you need these libaries for? If you need them for testing purpose only, import them as dev-dependencies: https://github.com/integritee-network/worker/blob/master/core-primitives/types/Cargo.toml#L39-L40

parentchain/src/lib.rs
haerdib: Unresolving this conversation, because it is not yet clear for me why you've decided to go for the last block hash instead of a map from `BlockNumber` -> `BlockHash.` Is there a specific reason?

parentchain/src/tests.rs
```diff
@@ -0,0 +1,56 @@
...
+use sp_runtime::{generic, traits::BlakeTwo256, DispatchError::BadOrigin};
+
+pub type Header = generic::Header<BlockNumber, BlakeTwo256>;
+
+#[test]
+fn verify_storage_works() {
+	let block_number = 3;
+	let parent_hash = H256::from_low_u64_be(420);
+
+	let header: Header = HeaderT::new(
```
haerdib, line 30: Did you notice that here you're using the Header Trait (`HeaderT::new()`) from `frame_support::sp_runtime`, whereas in the `non_root_account_errs` fn you're using `Header::new()` (i.e. the Header imported from `sp_runtime::generic`). Not that I mind, as it's for testing only 😋

haerdib (changes requested):
Looks good all in all, but it won't work yet due to std dependencies

parentchain/Cargo.toml
gaudenzkessler: dev-dependencies ❤️ 
And then wedo not need to set the std-feature?

haerdib: No. cargo test is std - only anyway and dev-deps will not be reimported.

parentchain/src/lib.rs
gaudenzkessler: Sorry I misunderstood. Why do we need a map? Why the simple storage is not enough?

clangenb: I don't think we'd want an ever-growing map in the pallet, this bloats the runtime storage indefinitely. If such a continuous mapping is wished for, this should be handled outside the runtime.

haerdib: Hm. The frame pallet does include this mapping. But agreed, it might be an overkill. Let's leave it as it  is because its only a template anyway.

gaudenzkessler: 🎉

haerdib (approved):
Looks good now :)

Issue #7, opened 2021-06-08
Extract typedefs to primitives


clangenb:
In the substraTEE-worker we have to import the whole runtime to merely access some types. We can also redefine them in the worker, but it would be best to have them in a separate crate.

clangenb:
While doing this, we can move the `Enclave`-builder implementation from `pallet-teerex` to either `test-utils` or primitives too.

haerdib:
This one has been closesd but I'd like to question it now: What's the use case of having a separate primitives crate containing all primitives compared to a primitives crate inside each pallet?

I.e. primitives/teerex vs teerex/primitives?

Because I don't see the use case of the first one. The use case of the second one would be in case someone would like to extract the teerex pallet only, only one folder (teerex) would need to be extracted. And the primitives can still be pulled separately.

clangenb:
I see your point. I think it depends on the use-case. In Encointer, all the pallets are closely entangled, so it's convenient for them to be in one place. They are even in a single crate there. I don't know about the pallet's introduced here.

However, it is not the goal that projects have a copy of the pallet-teerex. Ours should hopefully fit all purposes. Has this not been the case?

haerdib:
Polkadex has forked the pallet, though they didn't introduce new functionalities. I don't know about other forks, but I guess there always will be special use cases that might need something extraordinary. 

I don't think our pallets are that closely entangled.

clangenb:
Regardless of potential pallet forks, I agree that your design makes more sense. However, this should be in a follow-up issue: #18

PR #15, opened 2021-11-09
7/extract primitives

echevrier:
Fix issue #7


primitives/claims/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+
+*/
+
+//Primitives for claims
```
clangenb, line 18: Here you should use lib level doc-strings: `//!`

clangenb (approved):
Personally, I would have implemented the primitives as a single crate with various modules because most of the time we'd need all the primitives (this is at least what I think).
 
However, I think this is fine too. If we think that we need them altogether, we can simply write a `primitives` crate that re-exports all the sub-primitives crates.

echevrier:
> Personally, I would have implemented the primitives as a single crate with various modules because most of the time we'd need all the primitives (this is at least what I think).
> 
> However, I think this is fine too. If we think that we need them altogether, we can simply write a `primitives` crate that re-exports all the sub-primitives crates.

I separate them because I'm not sure that we/others will add all pallets. For ex: the teeracle pallet. So it's more modular, flexible

PR #19, opened 2021-11-24
Upgrade to substrate 7406442bea0194ffcafc4e8d48d895d4d8d11346

echevrier:
update rust tool chain
clean std features: cargo build --release --no-default-feature runs

### DIFF
--- a/claims/src/lib.rs
+++ b/claims/src/lib.rs
@@ -37,6 +37,10 @@ use sp_runtime::{
 	},
 };
 use sp_std::{fmt::Debug, prelude::*};
+#[cfg(not(feature = "std"))]
+extern crate alloc;
+#[cfg(not(feature = "std"))]
+use alloc::vec;
 
 type CurrencyOf<T> = <<T as Config>::VestingSchedule as VestingSchedule<
 	<T as frame_system::Config>::AccountId,
--- a/claims/src/lib.rs
+++ b/claims/src/lib.rs
@@ -36,11 +36,7 @@ use sp_runtime::{
 		InvalidTransaction, TransactionValidity, TransactionValidityError, ValidTransaction,
 	},
 };
-use sp_std::{fmt::Debug, prelude::*};
-#[cfg(not(feature = "std"))]
-extern crate alloc;
-#[cfg(not(feature = "std"))]
-use alloc::vec;
+use sp_std::{fmt::Debug, prelude::*, vec};
 
 type CurrencyOf<T> = <<T as Config>::VestingSchedule as VestingSchedule<
 	<T as frame_system::Config>::AccountId,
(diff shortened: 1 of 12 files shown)


claims/src/lib.rs
```diff
@@ -37,6 +37,10 @@ use sp_runtime::{
 	},
 };
 use sp_std::{fmt::Debug, prelude::*};
+#[cfg(not(feature = "std"))]
+extern crate alloc;
+#[cfg(not(feature = "std"))]
+use alloc::vec;
```
clangenb, line 43: `vec` is also available in sp_std

clangenb (commented):
Looks good, but I'd go for substrate's vec there.

primitives/teerex/src/lib.rs
```diff
@@ -48,9 +47,6 @@ impl<PubKey, Url> Enclave<PubKey, Url> {
 
 pub type ShardIdentifier = H256;
 
-// Disambiguate associated types
-pub type AccountId<T> = <T as frame_system::Config>::AccountId;
-
```
clangenb, line 53: maybe you don't need the frame-system dependency anymore.

teeracle/Cargo.toml
```diff
@@ -124,20 +124,22 @@ path = "../primitives/teeracle"
...
     'frame-system/std',
+    "pallet-teerex/std",
+    'scale-info/std',
     'sp-core/std',
     'sp-io/std',
     'sp-runtime/std',
     'sp-std/std',
     'substrate-fixed/std',
+    "teeracle-primitives/std",
+    "test-utils/std",
```
clangenb, line 138: Is test-utils used by the benchmark tests? otherwise we can change this dep to be a dev-dep without opting out of the default-features.

clangenb (commented):
I still discovered minor questions here. :)

teeracle/Cargo.toml
haerdib: Same for the # benchmarking stuff below?

clangenb: No, it is the other way around, benchmarking stuff needs to be a real dependency.

echevrier: test-utils used by the benchmark.

echevrier: Shall I set it under runtime-benchmarks and make it optional?

primitives/teerex/src/lib.rs
echevrier: Still needed I think

clangenb: I am quite sure it isn't.

echevrier: you're right

haerdib (approved):
LGTM.

clangenb (approved):
Looks good to me!

PR #24, opened 2021-12-01
Add whitelist of trusted exchange rate oracles 

echevrier:
Update of exchange rate can only be done by a trusted oracle

### DIFF
--- a/teeracle/src/lib.rs
+++ b/teeracle/src/lib.rs
@@ -31,6 +31,7 @@
 //!
 #![cfg_attr(not(feature = "std"), no_std)]
 pub use crate::weights::WeightInfo;
+use codec::Encode;
 pub use pallet::*;
 pub use substrate_fixed::types::U32F32;
 
@@ -53,10 +54,16 @@ pub mod pallet {
 		type WeightInfo: WeightInfo;
 	}
 
+	/// Exchange rates chain's cryptocurrency/currency
 	#[pallet::storage]
 	#[pallet::getter(fn exchange_rate)]
 	pub(super) type ExchangeRates<T> =
 		StorageMap<_, Blake2_128Concat, CurrencyString, ExchangeRate, ValueQuery>;
+
+	/// whitelist of trusted exchange rate oracles
+	#[pallet::storage]
+	#[pallet::getter(fn whitelist)]
+	pub(super) type Whitelist<T> = StorageValue<_, Vec<[u8; 32]>, ValueQuery>;
 
 	#[pallet::event]
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
@@ -76,6 +83,42 @@ pub mod pallet {
 
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
+		#[pallet::weight(<T as Config>::WeightInfo::add_to_whitelist())]
+		pub fn add_to_whitelist(origin: OriginFor<T>, mrenclave: [u8; 32]) -> DispatchResult {
+			ensure_root(origin)?;
+			if Self::is_whitelisted(mrenclave) {
+				log::info!("Oracle already in the whitelist");
+				return Ok(())
+			}
+
+			/*let count = Whitelist::<T>::decode_len().unwrap_or(0) as u32;
+			count
+				.checked_add(1)
+				.ok_or("[Teeracle]: Overflow adding new oracle to whitelist")?;*/
+
+			<Whitelist<T>>::append(mrenclave);
+			Ok(())
+		}
+		#[pallet::weight(<T as Config>::WeightInfo::remove_from_whitelist())]
+		pub fn remove_from_whitelist(origin: OriginFor<T>, mrenclave: [u8; 32]) -> DispatchResult {
+			ensure_root(origin)?;
+			if !Self::is_whitelisted(mrenclave) {
+				log::info!("Oracle is not int whitelist");
+				return Ok(())
+			}
+			Whitelist::<T>::mutate(|mrenclaves| {
+				mrenclaves.retain(|m| m.encode() != mrenclave.encode())
+			});
+			Ok(())
+		}
+
+		#[pallet::weight(<T as Config>::WeightInfo::clear_whitelist())]
+		pub fn clear_whitelist(origin: OriginFor<T>) -> DispatchResult {
+			ensure_root(origin)?;
+			<Whitelist<T>>::kill();
+			Ok(())
+		}
+
 		#[pallet::weight(<T as Config>::WeightInfo::update_exchange_rate())]
 		pub fn update_exchange_rate(
 			origin: OriginFor<T>,
@@ -95,6 +138,11 @@ pub mod pallet {
 			}
 			Ok(().into())
 		}
+	}
+}
+impl<T: Config> Pallet<T> {
+	fn is_whitelisted(mrenclave: [u8; 32]) -> bool {
+		Self::whitelist().iter().any(|m| m.encode() == mrenclave.encode())
 	}
 }
 
--- a/teeracle/src/lib.rs
+++ b/teeracle/src/lib.rs
@@ -87,6 +87,7 @@ pub mod pallet {
 		TooManyOracles,
 		NonWhitelistedOracle,
 		AlreadyWhitelistedOracle,
+		UntrustedOracle,
 	}
 
 	#[pallet::hooks]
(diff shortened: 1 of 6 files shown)


teeracle/src/lib.rs
```diff
@@ -51,31 +52,76 @@ pub mod pallet {
...
+			<Whitelist<T>>::try_append(mrenclave).map_err(|_| Error::<T>::TooManyOracles)?;
+			Self::deposit_event(Event::AddedToWhitelist(mrenclave));
+			Ok(())
+		}
+		#[pallet::weight(<T as Config>::WeightInfo::remove_from_whitelist())]
+		pub fn remove_from_whitelist(origin: OriginFor<T>, mrenclave: [u8; 32]) -> DispatchResult {
+			ensure_root(origin)?;
+			ensure!(Self::is_whitelisted(mrenclave), <Error<T>>::NonWhitelistedOracle);
+			Whitelist::<T>::mutate(|mrenclaves| {
+				mrenclaves.retain(|m| m.encode() != mrenclave.encode())
```
clangenb, line 111: I guess you can spare the encode here.

teeracle/src/lib.rs
```diff
@@ -97,6 +148,15 @@ pub mod pallet {
 		}
 	}
 }
+impl<T: Config> Pallet<T> {
+	fn is_whitelisted(mrenclave: [u8; 32]) -> bool {
+		Self::whitelist().iter().any(|m| m.encode() == mrenclave.encode())
+	}
```
clangenb, line 154: If it is possible, you can also drop the encode here.

teeracle/src/lib.rs
```diff
@@ -51,31 +52,76 @@ pub mod pallet {
...
-	pub(super) type ExchangeRates<T> =
+	pub(super) type ExchangeRates<T: Config> =
 		StorageMap<_, Blake2_128Concat, CurrencyString, ExchangeRate, ValueQuery>;
 
+	/// whitelist of trusted exchange rate oracles
+	#[pallet::storage]
+	#[pallet::getter(fn whitelist)]
+	pub(super) type Whitelist<T: Config> =
+		StorageValue<_, WeakBoundedVec<[u8; 32], T::MaxOracles>, ValueQuery>;
+
```
clangenb, line 71: I have never heard about this, and couldn't find material. Can you give me some background about the weakbounded vec? :)

teeracle/src/tests.rs
```diff
@@ -59,7 +74,8 @@ fn verifiy_update_exchange_rate_works() {
 fn verifiy_get_existing_exchange_rate_works() {
```
clangenb, line 74: I just noticed some typos in the verify. :)

clangenb: I am not fond of the verify prefix anyhow. I don't think it adds information, but only lengthens the function name unnecessarily.

clangenb (commented):
Minor comments, but looks good.

I like the cleanly written tests!

teeracle/src/lib.rs
echevrier: Without encode: no implementation for `&[u8; 32] == [u8; 32]`

clangenb: `|m| *m != mrenclave` should solve your problem then.

teeracle/src/lib.rs
clangenb: Found the info, thanks

echevrier: See: https://github.com/paritytech/substrate/blob/master/frame/support/src/storage/weak_bounded_vec.rs
This is similar to BoundedVec, but less strict. It allows some methods to go beyond the vector boundary,  logging in such cases logs error messages, but have methods like try-append, try_mutate, which ensure not to go beyond some length
I used it to ensure the limit of the whitelist size.

teeracle/src/lib.rs
echevrier: Ouch

teeracle/src/tests.rs
echevrier: OK. I remove then all

clangenb (approved):
Looks good to me, nice and clean.

teeracle/src/lib.rs
```diff
@@ -51,31 +51,74 @@ pub mod pallet {
...
 	}
 
 	#[pallet::error]
 	pub enum Error<T> {
 		InvalidCurrency,
+		/// Too many oracles in the whitelist.
+		TooManyOracles,
+		NonWhitelistedOracle,
+		AlreadyWhitelistedOracle,
+		UntrustedOracle,
```
brenzi, line 89: This is the same as NonWhitelistedOracle
please merge the two errors. The origin of the error is different, but the meaning is the same IMO

teeracle/src/lib.rs
```diff
@@ -51,31 +51,74 @@ pub mod pallet {
...
+		AddedToWhitelist([u8; 32]),
+		RemovedFromWhitelist([u8; 32]),
+		OracleWhitelistCleared,
 	}
 
 	#[pallet::error]
 	pub enum Error<T> {
 		InvalidCurrency,
+		/// Too many oracles in the whitelist.
+		TooManyOracles,
```
brenzi, line 86: 1. You are not whitelisting Oracles (service instances), you are whitelisting Enclaves (binaries). I suggest you refactor `Oracle` to `MrEnclave` here and everywhere else where it makes sense
2. These are all "whitelisting" errors, so I think they should all have "Whitelist" in their name

brenzi: suggestion:
```
MrEnclaveWhitelistOverflow
MrEnclaveNotWhitelisted
MrEnclaveAlreadyWhitelisted
```

teeracle/src/lib.rs
```diff
@@ -97,6 +145,15 @@ pub mod pallet {
 		}
 	}
 }
+impl<T: Config> Pallet<T> {
+	fn is_whitelisted(mrenclave: [u8; 32]) -> bool {
+		Self::whitelist().iter().any(|m| *m == mrenclave)
```
brenzi, line 150: why not `Self::whitelist().contains(mrenclave)` ? It's a vec, right?

teeracle/src/lib.rs
```diff
@@ -97,6 +145,15 @@ pub mod pallet {
...
+	}
+
+	pub fn whitelisted_oracle_count() -> u32 {
```
brenzi, line 153: why a helper function for this? the name is longer than the actual direct instruction

teeracle/src/mock.rs
```diff
@@ -120,6 +120,7 @@ impl timestamp::Config for Test {
 parameter_types! {
 	pub const MomentsPerDay: u64 = 86_400_000; // [ms/d]
 	pub const MaxSilenceTime: u64 = 172_800_000; // 48h
+	pub const MaxOracles: u32 = 2;
```
brenzi, line 123: rather `MaxWhitelistLength`?
and set it higher as 2 is very limiting for tests. maybe 10?

teeracle/src/lib.rs
```diff
@@ -51,31 +51,74 @@ pub mod pallet {
...
+		pub fn remove_from_whitelist(origin: OriginFor<T>, mrenclave: [u8; 32]) -> DispatchResult {
+			ensure_root(origin)?;
+			ensure!(Self::is_whitelisted(mrenclave), <Error<T>>::NonWhitelistedOracle);
+			Whitelist::<T>::mutate(|mrenclaves| mrenclaves.retain(|m| *m != mrenclave));
+			Self::deposit_event(Event::RemovedFromWhitelist(mrenclave));
+			Ok(())
+		}
+
+		#[pallet::weight(<T as Config>::WeightInfo::clear_whitelist())]
+		pub fn clear_whitelist(origin: OriginFor<T>) -> DispatchResult {
```
brenzi, line 115: what is this function good for? I don't see a reason to have it as we can remove all until empty one by one. It's just one more function to maintain

brenzi (changes requested):

teeracle/src/lib.rs
echevrier: I think these errors should be user-friendly. And I'm not sure that `MrEnclaveNotWhitelisted`, ... is clear for a dummy user.
From a user's perspective, is it wrong to say it is about the whitelisting of oracles?  In the issue description, you wrote "_the teeracle should maintain a whitelist of oracles_".

teeracle/src/mock.rs
echevrier: This is a param that has to be configured in the Node runtime. Here again, I think that MaxWhitelistLength is not so "user-friendly". If "oracle" is ok, then what about MaxTrustedOracles ?

teeracle/src/lib.rs
brenzi: It seems I wasn't precise. What we are whitelisting is MrEnclave. But it is true that MrEnclave is very SGX specific and not generic for all TEE technologies.
What about `Release`? Each oracle service runs a specific release. And MrEnclave identifies a release unambiguously

echevrier: OK

teeracle/src/mock.rs
echevrier: What about this?

brenzi: MaxWhitelistedReleases

brenzi (approved):

PR #26, opened 2021-12-21
Teeracle support multiple sources and multiple trading pairs

echevrier:
Whitelist is a storage map. Key: source, Value: WeakBoundedVec of mrenclaves . A mrenclave in the list can operate the source.
ExchangeRates is a storage double map: Key1: the trading pair, Key2: the source, Value the exchange rate.

### DIFF
--- a/primitives/common/Cargo.toml
+++ b/primitives/common/Cargo.toml
@@ -6,7 +6,6 @@ homepage = "https://integritee.network/"
 repository = "https://github.com/integritee-network/pallets/"
 license = "Apache-2.0"
 edition = "2021"
-resolver = "2"
 
 [dependencies]
 
--- a/primitives/common/Cargo.toml
+++ b/primitives/common/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "common-primitives"
+version = "0.1.0"
+authors = ["Integritee AG"]
+homepage = "https://integritee.network/"
+repository = "https://github.com/integritee-network/pallets/"
+license = "Apache-2.0"
+edition = "2021"
+resolver = "2"
+
+[dependencies]
+
+# substrate deps
+sp-std = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
+
+[features]
+default = ['std']
+std = [
+    'sp-std/std',
+]
--- a/primitives/common/src/lib.rs
+++ b/primitives/common/src/lib.rs
@@ -14,7 +14,7 @@
 	limitations under the License.
 
 */
-
+#![cfg_attr(not(feature = "std"), no_std)]
 //!Primitives for all pallets
 
 #[cfg(not(feature = "std"))]
--- a/primitives/common/src/lib.rs
+++ b/primitives/common/src/lib.rs
@@ -1,0 +1,46 @@
+/*
+	Copyright 2021 Integritee AG and Supercomputing Systems AG
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+		http://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+
+*/
+
+//!Primitives for all pallets
+
+#[cfg(not(feature = "std"))]
+use sp_std::vec::Vec;
+
+/// Substrate runtimes provide no string type. Hence, for arbitrary data of varying length the
+/// `Vec<u8>` is used. In the polkadot-js the typedef `Text` is used to automatically
+/// utf8 decode bytes into a string.
+#[cfg(not(feature = "std"))]
+pub type PalletString = Vec<u8>;
+
+#[cfg(feature = "std")]
+pub type PalletString = String;
+
+pub trait AsByteOrNoop {
+	fn as_bytes_or_noop(&self) -> &[u8];
+}
+
+impl AsByteOrNoop for PalletString {
+	#[cfg(feature = "std")]
+	fn as_bytes_or_noop(&self) -> &[u8] {
+		self.as_bytes()
+	}
+
+	#[cfg(not(feature = "std"))]
+	fn as_bytes_or_noop(&self) -> &[u8] {
+		self
+	}
+}
--- a/primitives/teeracle/src/lib.rs
+++ b/primitives/teeracle/src/lib.rs
@@ -1,7 +1,6 @@
 //!Primitives for teeracle
 #![cfg_attr(not(feature = "std"), no_std)]
 use common_primitives::PalletString;
-use sp_std::prelude::*;
 use substrate_fixed::types::U32F32;
 
 pub type ExchangeRate = U32F32;
--- a/primitives/teeracle/src/lib.rs
+++ b/primitives/teeracle/src/lib.rs
@@ -1,8 +1,9 @@
 //!Primitives for teeracle
 #![cfg_attr(not(feature = "std"), no_std)]
+use common_primitives::PalletString;
 use sp_std::prelude::*;
 use substrate_fixed::types::U32F32;
 
 pub type ExchangeRate = U32F32;
-pub type TradingPairString = Vec<u8>;
-pub type MarketDataSourceString = Vec<u8>;
+pub type TradingPairString = PalletString;
+pub type MarketDataSourceString = PalletString;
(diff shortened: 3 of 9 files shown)


primitives/teeracle/src/lib.rs
```diff
@@ -4,4 +4,5 @@ use sp_std::prelude::*;
 use substrate_fixed::types::U32F32;
 
 pub type ExchangeRate = U32F32;
-pub type CurrencyString = Vec<u8>;
+pub type TradinPairString = Vec<u8>;
```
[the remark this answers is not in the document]
brenzi: typo: TradingPairString

teeracle/src/benchmarking.rs
```diff
@@ -44,8 +44,9 @@ benchmarks! {
 		ensure_not_skipping_ra_check();
 		timestamp::Pallet::<T>::set_timestamp(TEST4_SETUP.timestamp.checked_into().unwrap());
 		let signer: T::AccountId = get_signer(TEST4_SETUP.signer_pub);
-		let currency = "usd".as_bytes().to_owned();
+		let trading_pair =  "DOT/USD".as_bytes().to_owned();
 		let rate = U32F32::from_num(43.65);
+		let data_source = "https://api.coingecko.com".as_bytes().to_owned();
```
brenzi, line 49: either you limit the size of these vec arguments in the dispatchable function or you benchmark with variable size of vecs such that longer vecs cost more.
As it is, these arguments are an attack vector

teeracle/src/lib.rs
```diff
@@ -93,52 +109,86 @@ pub mod pallet {
...
 		#[pallet::weight(<T as Config>::WeightInfo::update_exchange_rate())]
 		pub fn update_exchange_rate(
 			origin: OriginFor<T>,
-			currency: CurrencyString,
+			data_source: MarketDataSourceString,
+			trading_pair: TradinPairString,
 			new_value: Option<ExchangeRate>,
 		) -> DispatchResultWithPostInfo {
 			let sender = ensure_signed(origin)?;
 			<pallet_teerex::Module<T>>::is_registered_enclave(&sender)?;
```
brenzi, line 155: here you could ensure vec argument size limits (or benchmark the size as stated above)

brenzi (changes requested):
I think the easiest fix would be to limit trading pairs to 11 chars and url to 40 chars by ensureing argument length in the dispatchable. Doesn't make sense to accept arbitrary size arguments

teeracle/src/tests.rs
```diff
@@ -25,12 +25,18 @@ use test_utils::ias::consts::{
 	TEST5_SIGNER_PUB, TEST8_MRENCLAVE, URL,
 };
 
+const COINGECKO_SRC: &[u8] = "https://api.coingecko.com".as_bytes();
+const COINMARKETCAP_SRC: &[u8] = "https://coinmarketcap.com/".as_bytes();
+
+const DOT_USD_TRADING_PAIR: &[u8] = "DOT/USD".as_bytes();
+const TEER_USD_TRADING_PAIR: &[u8] = "TEER/USD".as_bytes();
+
```
clangenb, line 33: You can save quite some `to_owned`s below if you declare those as `Vec<u8>`

teeracle/src/benchmarking.rs
clangenb: Exactly. I guess it is more reasonable to limit the vec size.

clangenb (commented):
Looks  quite good already. Only some minor comments.

teeracle/src/tests.rs
echevrier: There are constants, I don't know how I can declare them as Vec<u8>

clangenb: Like this: `const COINGECKO_SRC: Vec<u8> = "https://api.coingecko.com".as_bytes().to_vec();`

clangenb: btw. `to_owned` is the same as `to_vec` for `&[u8]`

echevrier: to_owned or to_vec throw an error for constant:
 `calls in constants are limited to constant functions, tuple structs and tuple variants`.
 as const it can be &[u8] or &str, but not Vec or String

clangenb: Ah ok I see. Makes sense. But const can be vecs. The problem is that `to_vec` is not a `const fn`. But there is no way to create a non-empty vec with `const fn`s as it seems. :)

teeracle/src/tests.rs
```diff
@@ -25,12 +25,18 @@ use test_utils::ias::consts::{
...
+const DOT_USD_TRADING_PAIR: &str = "DOT/USD";
+const TEER_USD_TRADING_PAIR: &str = "TEER/USD";
+
 // give get_signer a concrete type
 fn get_signer(pubkey: &[u8; 32]) -> AccountId {
 	test_utils::get_signer(pubkey)
 }
 
-fn register_enclave_and_add_oracle_to_whitelist_ok() {
+fn register_enclave_and_add_oracle_to_whitelist_ok(src: &str) {
```
clangenb, line 39: beautiful

primitives/common/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+	#[cfg(feature = "std")]
+	fn as_bytes_or_noop(&self) -> &[u8] {
+		self.as_bytes()
+	}
+
+	#[cfg(not(feature = "std"))]
+	fn as_bytes_or_noop(&self) -> &[u8] {
+		self
+	}
+}
```
clangenb, line 46: Nice, I forgot that you need to add these.

primitives/common/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "common-primitives"
+version = "0.1.0"
+authors = ["Integritee AG"]
+homepage = "https://integritee.network/"
+repository = "https://github.com/integritee-network/pallets/"
+license = "Apache-2.0"
+edition = "2021"
+resolver = "2"
```
clangenb, line 9: Resolver 2 is no longer needed. It is the default in edition 2021

teeracle/src/lib.rs
```diff
@@ -93,52 +114,91 @@ pub mod pallet {
...
+				),
 				<Error<T>>::ReleaseNotWhitelisted
 			);
 			if new_value.is_none() || new_value == Some(U32F32::from_num(0)) {
 				log::info!("Delete exchange rate : {:?}", new_value);
-				ExchangeRates::<T>::mutate_exists(currency.clone(), |rate| *rate = None);
-				Self::deposit_event(Event::ExchangeRateDeleted(currency));
+				ExchangeRates::<T>::mutate_exists(
+					trading_pair.clone(),
+					data_source.clone(),
```
clangenb, line 178: I think you should be able to use `&trading_pair` and remove the clone

teeracle/src/lib.rs
```diff
@@ -93,52 +114,91 @@ pub mod pallet {
...
 			}
 			Ok(().into())
 		}
 	}
 }
 impl<T: Config> Pallet<T> {
-	fn is_whitelisted(mrenclave: [u8; 32]) -> bool {
-		Self::whitelist().contains(&mrenclave)
+	fn is_whitelisted(data_source: MarketDataSourceString, mrenclave: [u8; 32]) -> bool {
+		Self::whitelist(data_source).contains(&mrenclave)
```
clangenb, line 201: Here you should also be able to use `&data_source` and thereby pass `&MarketDataSourceString`. This will spare you a bunch of clones.

clangenb (commented):
Found some other minor comments, but it looks good!

brenzi (approved):

clangenb (approved):
I will also approve, so you can merge after fixing/looking at my comments

teeracle/src/lib.rs
```diff
@@ -93,52 +114,85 @@ pub mod pallet {
...
+		) -> DispatchResult {
 			ensure_root(origin)?;
-			ensure!(Self::is_whitelisted(mrenclave), <Error<T>>::ReleaseNotWhitelisted);
-			Whitelist::<T>::mutate(|mrenclaves| mrenclaves.retain(|m| *m != mrenclave));
-			Self::deposit_event(Event::RemovedFromWhitelist(mrenclave));
+			ensure!(
+				Self::is_whitelisted(&data_source, mrenclave),
+				<Error<T>>::ReleaseNotWhitelisted
+			);
+			<Whitelists<T>>::mutate(data_source.clone(), |mrenclave_vec| {
```
clangenb, line 146: You forgot one ;)

echevrier: Ouch

primitives/common/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+
+*/
+
+//!Primitives for all pallets
+
```
clangenb, line 19: You are missing the following line here: `#![cfg_attr(not(feature = "std"), no_std)]`

PR #27, opened 2021-12-22
Fix benchmarks in no_std

clangenb:


teeracle/src/benchmarking.rs
```diff
@@ -44,9 +45,9 @@ benchmarks! {
 		ensure_not_skipping_ra_check();
 		timestamp::Pallet::<T>::set_timestamp(TEST4_SETUP.timestamp.checked_into().unwrap());
 		let signer: T::AccountId = get_signer(TEST4_SETUP.signer_pub);
-		let trading_pair =  "DOT/USD".to_owned();
+		let trading_pair: TradingPairString =  "DOT/USD".to_owned().into();
```
clangenb, line 48: Annotate the type explicitly to help the compiler with type inference. `into` has the nice characteristic that is simply does nothing if the type already matches.

echevrier (approved):
Cool thanks.

PR #37, opened 2022-02-01
Update pallet teerex to frame v2

haerdib:
Metadata of pallet did not change at all.

### DIFF
--- a/teerex/src/lib.rs
+++ b/teerex/src/lib.rs
@@ -123,13 +123,7 @@ pub mod pallet {
 	#[pallet::genesis_build]
 	impl<T: Config> GenesisBuild<T> for GenesisConfig {
 		fn build(&self) {
-			{
-				let data = &self.allow_sgx_debug_mode;
-				let v: &bool = data;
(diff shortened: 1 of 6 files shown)


teerex/src/lib.rs
```diff
@@ -40,72 +39,130 @@ use ias_verify::SgxBuildMode;
...
+	#[pallet::genesis_build]
+	impl<T: Config> GenesisBuild<T> for GenesisConfig {
+		fn build(&self) {
+			{
+				let data = &self.allow_sgx_debug_mode;
+				let v: &bool = data;
+				<AllowSGXDebugMode<T> as frame_support::storage::StorageValue<bool>>::put::<&bool>(
+					v,
+				);
+			}
```
clangenb, line 132: This one is unnecessarily verbose due to the migration template, you can simplify this to a one liner:

```
<AllowSGXDebugMode<T>>::put(&self.allow_sgx_debug_mode);

```

clangenb (commented):
Looks good to me, one style comment.

teerex/src/lib.rs
haerdib: Good catch, updated accordingly

clangenb (approved):
Coolio

PR #47, opened 2022-04-29
Sidechain block finalization

gaudenzkessler:
* check that block number is consecutive to the previous one
 * check that block hash equals the hash of the previous block

related to issue https://github.com/integritee-network/integritee-node/issues/115

### DIFF
--- a/primitives/teerex/Cargo.toml
+++ b/primitives/teerex/Cargo.toml
@@ -16,6 +16,7 @@ serde = { version = "1.0.130", default-features = false }
 ias-verify = { path = "../../teerex/ias-verify", default-features = false }
 
 # substrate dependencies
+sp-io = { version = "6.0.0", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.20" }
 sp-core = { version = "6.0.0", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.20" }
 sp-std = { version = "4.0.0-dev", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.20" }
 
--- a/primitives/teerex/Cargo.toml
+++ b/primitives/teerex/Cargo.toml
@@ -10,6 +10,7 @@ edition = "2021"
 [dependencies]
 codec = { version = "3.0.0", default-features = false, features = ["derive"], package = "parity-scale-codec" }
 scale-info = { version = "2.0.1", default-features = false, features = ["derive"] }
+serde = { version = "1.0.130", default-features = false }
 
 # local
 ias-verify = { path = "../../teerex/ias-verify", default-features = false }
@@ -24,6 +25,7 @@ default = ["std"]
 std = [
     "codec/std",
     "scale-info/std",
+    "serde/std",
     # local
     "ias-verify/std",
     # substrate
--- a/primitives/teerex/Cargo.toml
+++ b/primitives/teerex/Cargo.toml
@@ -31,5 +31,6 @@ std = [
     "ias-verify/std",
     # substrate
     "sp-core/std",
+    "sp-io/std",
     "sp-std/std",
 ]
--- a/primitives/teerex/src/lib.rs
+++ b/primitives/teerex/src/lib.rs
@@ -23,6 +23,9 @@ use scale_info::TypeInfo;
 use sp_core::H256;
 use sp_std::prelude::*;
 
+#[cfg(feature = "std")]
+use serde::{Deserialize, Serialize};
+
 #[derive(Encode, Decode, Default, Copy, Clone, PartialEq, sp_core::RuntimeDebug, TypeInfo)]
 pub struct Enclave<PubKey, Url> {
 	pub pubkey: PubKey, // FIXME: this is redundant information
@@ -51,4 +54,20 @@ pub type ShardIdentifier = H256;
 pub struct Request {
 	pub shard: ShardIdentifier,
 	pub cyphertext: Vec<u8>,
+}
+
+#[derive(PartialEq, Eq, Clone, Encode, Decode, Debug, Copy, Default, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct Header {
+	/// The parent hash.
+	pub parent_hash: H256,
+
+	/// The block number.
+	pub block_number: u64,
+
+	/// The Shard id.
+	pub shard_id: ShardIdentifier,
+
+	/// The payload hash.
+	pub block_data_hash: H256,
 }
--- a/primitives/teerex/src/lib.rs
+++ b/primitives/teerex/src/lib.rs
@@ -59,7 +59,7 @@ pub struct Request {
 
 #[derive(PartialEq, Eq, Clone, Encode, Decode, Debug, Copy, Default, TypeInfo)]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
-pub struct Header {
+pub struct SidechainHeader {
 	/// The parent hash.
 	pub parent_hash: H256,
 
@@ -73,7 +73,7 @@ pub struct Header {
 	pub block_data_hash: H256,
 }
 
-impl Header {
+impl SidechainHeader {
 	/// get the `blake2_256` hash of the header.
 	pub fn hash(&self) -> H256 {
 		self.using_encoded(blake2_256).into()
--- a/primitives/teerex/src/lib.rs
+++ b/primitives/teerex/src/lib.rs
@@ -21,6 +21,7 @@ use codec::{Decode, Encode};
 use ias_verify::SgxBuildMode;
 use scale_info::TypeInfo;
 use sp_core::H256;
+use sp_io::hashing::blake2_256;
 use sp_std::prelude::*;
 
 #[cfg(feature = "std")]
@@ -70,4 +71,11 @@ pub struct Header {
 
 	/// The payload hash.
 	pub block_data_hash: H256,
+}
+
+impl Header {
+	/// get the `blake2_256` hash of the header.
+	pub fn hash(&self) -> H256 {
+		self.using_encoded(blake2_256).into()
+	}
 }
--- a/teerex/src/lib.rs
+++ b/teerex/src/lib.rs
@@ -111,11 +111,19 @@ pub mod pallet {
 
 	#[pallet::storage]
 	#[pallet::getter(fn latest_sidechain_block_header)]
-	pub type LatestSidechainBlockHeader<T: Config> = StorageValue<_, Header, ValueQuery>;
+	pub type LatestSidechainBlockHeader<T: Config> =
+		StorageMap<_, Blake2_128Concat, ShardIdentifier, Header, ValueQuery>;
 
 	#[pallet::storage]
-	pub type SidechainBlockHeaderQueue<T: Config> =
-		StorageMap<_, Blake2_128Concat, u64, Header, ValueQuery>;
+	pub type SidechainBlockHeaderQueue<T: Config> = StorageDoubleMap<
+		_,
+		Blake2_128Concat,
+		ShardIdentifier,
+		Blake2_128Concat,
+		u64,
+		Header,
+		ValueQuery,
+	>;
 
 	#[pallet::storage]
 	#[pallet::getter(fn allow_sgx_debug_mode)]
@@ -250,7 +258,7 @@ pub mod pallet {
 			);
 
 			let early_block_proposal_lenience = T::EarlyBlockProposalLenience::get();
-			let mut latest_block_header = <LatestSidechainBlockHeader<T>>::get();
+			let mut latest_block_header = <LatestSidechainBlockHeader<T>>::get(shard_id);
 			let mut latest_block_number = latest_block_header.block_number;
 			let block_number = block_header.block_number;
 
@@ -261,8 +269,8 @@ pub mod pallet {
 			{
 				return Err(<Error<T>>::BlockNumberTooHigh.into())
 			} else if block_number > latest_block_number + 1 {
-				if !<SidechainBlockHeaderQueue<T>>::contains_key(block_number) {
-					<SidechainBlockHeaderQueue<T>>::insert(block_number, block_header);
+				if !<SidechainBlockHeaderQueue<T>>::contains_key(shard_id, block_number) {
+					<SidechainBlockHeaderQueue<T>>::insert(shard_id, block_number, block_header);
 				}
 			} else if block_number ==
 				latest_block_number
@@ -278,11 +286,14 @@ pub mod pallet {
 				);
 				latest_block_number = latest_block_header.block_number;
 				let mut i: u64 = 0;
-				while <SidechainBlockHeaderQueue<T>>::contains_key(latest_block_number + 1) &&
-					i < early_block_proposal_lenience
+				while <SidechainBlockHeaderQueue<T>>::contains_key(
+					shard_id,
+					latest_block_number + 1,
+				) && i < early_block_proposal_lenience
 				{
-					let header = <SidechainBlockHeaderQueue<T>>::take(latest_block_number + 1);
-					<SidechainBlockHeaderQueue<T>>::remove(latest_block_number + 1);
+					let header =
+						<SidechainBlockHeaderQueue<T>>::take(shard_id, latest_block_number + 1);
+					<SidechainBlockHeaderQueue<T>>::remove(shard_id, latest_block_number + 1);
 					latest_block_header = Self::check_block_and_get_latest_header(
 						shard_id,
 						header,
@@ -528,7 +539,7 @@ impl<T: Config> Pallet<T> {
 		sender: &T::AccountId,
 		sender_index: u64,
 	) {
-		<LatestSidechainBlockHeader<T>>::put(block_header);
+		<LatestSidechainBlockHeader<T>>::insert(shard_id, block_header);
 		<WorkerForShard<T>>::insert(shard_id, sender_index);
 		let block_hash = block_header.block_data_hash;
 		log::debug!(
--- a/teerex/src/lib.rs
+++ b/teerex/src/lib.rs
@@ -107,6 +107,14 @@ pub mod pallet {
 	pub type ExecutedCalls<T: Config> = StorageMap<_, Blake2_128Concat, H256, u64, ValueQuery>;
 
 	#[pallet::storage]
+	#[pallet::getter(fn lastest_sidechain_block_header)]
+	pub type LatestSidechainBlockHeader<T: Config> = StorageValue<_, Header, ValueQuery>;
+
+	#[pallet::storage]
+	pub type SidechainBlockHeaderQueue<T: Config> =
+		StorageMap<_, Blake2_128Concat, u64, Header, ValueQuery>;
+
+	#[pallet::storage]
 	#[pallet::getter(fn allow_sgx_debug_mode)]
 	pub type AllowSGXDebugMode<T: Config> = StorageValue<_, bool, ValueQuery>;
 
@@ -226,7 +234,7 @@ pub mod pallet {
 		pub fn confirm_proposed_sidechain_block(
 			origin: OriginFor<T>,
 			shard_id: ShardIdentifier,
-			block_hash: H256,
+			block_header: Header,
 		) -> DispatchResultWithPostInfo {
 			let sender = ensure_signed(origin)?;
 			Self::is_registered_enclave(&sender)?;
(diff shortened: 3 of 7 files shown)


teerex/src/lib.rs
```diff
@@ -237,13 +245,40 @@ pub mod pallet {
...
+			log::error!("Got new block: {:?}", block_number);
+			if latest_block_number + 1 == block_number {
+				latest_block_header = Self::check_block_and_get_latest_header(
+					shard_id,
+					block_header,
+					&sender,
+					sender_index,
+					latest_block_header,
+				);
+			} else if latest_block_number + 1 < block_number {
```
mullefel, line 261: What happens if `block_number < latest_block_number + 1` ? It seems we then still set `latest_block_number = latest_block_header.block_number`, which is probably not correct then? In that case we should just discard the header and do nothing, I believe?

primitives/teerex/src/lib.rs
```diff
@@ -52,3 +56,26 @@ pub struct Request {
...
+
+	/// The block number.
+	pub block_number: u64,
+
+	/// The Shard id.
+	pub shard_id: ShardIdentifier,
+
+	/// The payload hash.
+	pub block_data_hash: H256,
+}
```
mullefel, line 74: Is this a pattern that we have for other data structures as well, i.e. duplicating code of the worker?

teerex/src/lib.rs
gaudenzkessler: Yes, that is what I understood, that we discard headers with the same numer that are already imported.

teerex/src/lib.rs
```diff
@@ -446,6 +481,45 @@ impl<T: Config> Pallet<T> {
...
+		log::error!("Confirming block: {:?}", block_header.block_number);
+		<LatestSidechainBlockHeader<T>>::put(block_header);
+		<WorkerForShard<T>>::insert(shard_id, sender_index);
+		let block_hash = block_header.block_data_hash;
+		log::debug!(
+			"Proposed sidechain block confirmed with shard {:?}, block hash {:?}",
+			shard_id,
+			block_hash
+		);
+		Self::deposit_event(Event::ProposedSidechainBlock(sender.clone(), block_hash));
```
mullefel, line 521: How are sidechain blocks finalized in this way? Are they included in the parentchain block somehow or not? Does an event mean we have to subscribe to these events to get the information, or can we also get the information from a parentchain block?

teerex/src/lib.rs
mullefel: Okay, but that's not what happens in the code as I see it here?

gaudenzkessler: yes, we chan skip the rest of the method for  latest_block_number >= block_number

teerex/src/lib.rs
```diff
@@ -106,6 +106,14 @@ pub mod pallet {
 	#[pallet::getter(fn confirmed_calls)]
 	pub type ExecutedCalls<T: Config> = StorageMap<_, Blake2_128Concat, H256, u64, ValueQuery>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn latest_sidechain_block_header)]
+	pub type LatestSidechainBlockHeader<T: Config> = StorageValue<_, Header, ValueQuery>;
+
+	#[pallet::storage]
+	pub type SidechainBlockHeaderQueue<T: Config> =
+		StorageMap<_, Blake2_128Concat, u64, Header, ValueQuery>;
```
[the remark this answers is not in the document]
clangenb: Also, we need to have a mechanism that prevents orphan blocks in the hashmap that will never be removed. I am not yet sure how to solve this, so I made an issue for that: https://github.com/integritee-network/pallets/issues/48

teerex/src/lib.rs
clangenb: Additionally, I want to have a safety-param here: 

```
latest_block_number + 1 < block_number + EARLY_BLOCK_PROPOSAL_LENIENCE
```

To prevent the arrival of blocks that are far too early. If this check fails, an error should be returned.

This safety param should be introduced as a runtime-constant.

teerex/src/lib.rs
```diff
@@ -237,13 +245,37 @@ pub mod pallet {
...
+			if latest_block_number + 1 == block_number {
+				latest_block_header = Self::check_block_and_get_latest_header(
+					shard_id,
+					block_header,
+					&sender,
+					sender_index,
+					latest_block_header,
+				);
+				latest_block_number = latest_block_header.block_number;
+				while <SidechainBlockHeaderQueue<T>>::contains_key(latest_block_number + 1) {
```
clangenb, line 261: while loops in a blockchain dispatchable are very dangerous without a formal proof that this will not run for too long, this is not acceptable.

I want to incorporate here the `EARLY_BLOCK_PROPOSAL_LENIENCE` as an upper bound for iterations. (See comment below for that).

teerex/src/lib.rs
clangenb: I have quickly spoken with @gaudenzkessler regarding this. I currently believe it is enough to just store the latest one. The worker shall find out on its own if it is not on the canonical chain, and fetch the correct blocks from another worker.

clangenb (changes requested):

primitives/teerex/src/lib.rs
haerdib: Yes, we currently are duplicating a lot. Before bickering about this extra duplication, we should define where we would want a shared primitives crate: #50

teerex/src/lib.rs
mullefel: That means however, that we roll-back a lot of sidechain blocks in case of a fork. If we only know the last finalized block, we don't know the exact block where the fork happened, and we need to roll-back everything until the last known finalized block. Which will be in the order of 40 sidechain blocks. E.g. if the fork happens at sidechain block 38 and we did know all the finalized blocks, we only roll back 2, not 40.

haerdib: The worker picks up every extrinsic that is included in a parentchain block, in this case that would be the `confirm_proposed_sidechain_block(origin: OriginFor<T>, shard_id: ShardIdentifier, block_header: Header)` extrinsic. So the worker itself knows the if an extrinsic with a specific header has been successful or not. It does not need the parentchain state for this.

Or am I missing something?

mullefel: But it needs to know if a sidechain block was finalized, not if an extrinsic was successful or not? An extrinsic can be successful, but still be on a sidechain block that will not be finalized.

teerex/src/lib.rs
haerdib: I honestly don't like the approach of "doing nothing". This is hell for every developer trying to debug. Can't you return an error instead?

teerex/src/lib.rs
haerdib: Ah, I see. So it's also successful if it's added into the queue. Hm. Then that doesn't work, my bad.

But lets say: The xt is successful if, and only if, the header gets added to the queue or is directly confirmed. And a header gets removed from the queue once it's been confirmed.

So depending on how we solve #48, can't we deduct the finalized headers by looking up the queue at the current parentchain block state, take all headers that have been successfully included in the block with `confirm_proposed_sidechain_block` minus the ones left in queue?

mullefel: Maybe we need to have a separate meeting for this, I think it's too much to discuss in the comments here. I believe the concept of `confirm_proposed_sidechain_block` will need to change or go away in the future, because it's not compatible with the finalization mechanism. When a sidechain block is proposed, even if it's perfectly valid from a worker perspective, it does not mean that it gets finalized by the parentchain.

teerex/src/lib.rs
mullefel: But it is not an error? When we encounter such a header, we need to discard/ignore it. This is not an error case, it's part of the regular operation of this function. In addition to testing (which demonstrates the mechanism), we could add a `debug!()` or `trace!()` log.

haerdib: Let's schedule a meeting. :)

teerex/src/lib.rs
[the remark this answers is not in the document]
gaudenzkessler: should we just purge the queue after a new block is imported?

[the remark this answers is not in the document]
clangenb: We could do that, but we should not purge the entire queue, but only up to `latest_sidechain_header.numer`, no?

teerex/src/lib.rs
```diff
@@ -258,9 +285,15 @@ pub mod pallet {
...
-					let header = <SidechainBlockHeaderQueue<T>>::take(latest_block_number + 1);
-					<SidechainBlockHeaderQueue<T>>::remove(latest_block_number + 1);
+				let mut i: u64 = 0;
+				while <SidechainBlockHeaderQueue<T>>::contains_key(
+					shard_id,
+					latest_block_number + 1,
+				) && i < early_block_proposal_lenience
+				{
+					let header =
+						<SidechainBlockHeaderQueue<T>>::take(shard_id, latest_block_number + 1);
```
clangenb, line 295: I mean theoretically, we checked before that `latest_block_number + 1` does not overflow, but I'd still like to see the checked operations. Such a pattern is not safe when aposterio code changes remove the checked math operatios a few lines above.

Anyhow, here we could just allocate a variable `next_block_number` to produce less overhead code.

clangenb (commented):
Minor remarks looks good otherwise

teerex/src/lib.rs
```diff
@@ -361,6 +393,10 @@ pub mod pallet {
 		RaReportTooLong,
 		/// No enclave is registered.
 		EmptyEnclaveRegistry,
+		/// A proposed block is too early.
+		BlockNumberTooHigh,
+		/// bla
```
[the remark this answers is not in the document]
gaudenzkessler: ooops 😊 👼

teerex/src/lib.rs
gaudenzkessler: Let's return an error

gaudenzkessler: We decided to return the errors
 * BlockNumberTooHigh in case the blocknumber is lenience above the latest blocknumber and
 * OutdatedBlockNumber in case the blocknumber is below the latest blocknumber.

teerex/src/lib.rs
gaudenzkessler: We decided to keep the latest block in order to keep the chain state small

teerex/src/lib.rs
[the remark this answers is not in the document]
gaudenzkessler: For this algorithm a purge of the queue should not be required.

teerex/src/lib.rs
gaudenzkessler: I created a function to compare with the blocknumber, since the code is also very similar to the lenience comparison.

primitives/teerex/src/lib.rs
```diff
@@ -52,3 +56,26 @@ pub struct Request {
 	pub shard: ShardIdentifier,
 	pub cyphertext: Vec<u8>,
 }
+
+#[derive(PartialEq, Eq, Clone, Encode, Decode, Debug, Copy, Default, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct Header {
```
haerdib, line 62:
```suggestion
pub struct SidechainBlockHeader {
```
Otherwise, things might get very confusing

teerex/src/lib.rs
```diff
@@ -106,6 +109,24 @@ pub mod pallet {
 	#[pallet::getter(fn confirmed_calls)]
 	pub type ExecutedCalls<T: Config> = StorageMap<_, Blake2_128Concat, H256, u64, ValueQuery>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn latest_sidechain_block_header)]
+	pub type LatestSidechainBlockHeader<T: Config> =
+		StorageMap<_, Blake2_128Concat, ShardIdentifier, Header, ValueQuery>;
+
+	pub type ShardBlockNumber = (ShardIdentifier, u64);
```
haerdib, line 117: Can you move the type defintion above `LatestSidechainBlockHeader` ? Substrate convention is that storage, calls and such are always grouped together, which makes the large files a little less messy.

teerex/src/lib.rs
```diff
@@ -446,6 +500,78 @@ impl<T: Config> Pallet<T> {
...
+		let block_hash = block_header.block_data_hash;
+		log::debug!(
+			"Proposed sidechain block confirmed with shard {:?}, block hash {:?}",
+			shard_id,
+			block_hash
+		);
+		Self::deposit_event(Event::ProposedSidechainBlock(sender.clone(), block_hash));
+	}
+
+	fn import_blocks_from_queue(
```
mullefel, line 538: I think `import` is not the correct term here? We finalize the blocks, not import.

teerex/src/lib.rs
```diff
@@ -237,13 +258,42 @@ pub mod pallet {
...
-				block_hash
-			);
-			Self::deposit_event(Event::ProposedSidechainBlock(sender, block_hash));
+
+			let lenience = T::EarlyBlockProposalLenience::get();
+			let mut latest_block_header = <LatestSidechainBlockHeader<T>>::get(shard_id);
+			let latest_block_number = latest_block_header.block_number;
+			let block_number = block_header.block_number;
+
+			if block_number > Self::compare_block_number(latest_block_number, lenience)? {
```
haerdib, line 267: `compare_block_number` does not compare but add the numbers. But before changing this, I'd suggest a totally different  design altogether. Because it is hard to understand why the same function is called three times.. the differing 2nd input variable in only visible when looking at veery intensely.

Suggestion:

```rust
ensure!(block_number > latest_block_number , Error::<T>::OutdatedBlockNumber)
let MAX_ALLOWED_BLOCK_NUMBER = latest_block_number.checked_add(1).ok_or(Error::<T>::CheckedAddOverflow)?
ensure!(block_number < MAX_ALLOWED_BLOCK_NUMBER , Error::<T>::BlockNumberTooHigh)

let expected_block_number = latest_block_number.checked_add(1).ok_or(Error::<T>::CheckedAddOverflow)?
if block_number == expected_block_number {
 Self::finalize_sidechain_block(block_header)?
 Self::import_blocks_from_queue(...)?
} else {
 Self::import_to_queue(block_header)?;
}
Ok(())
```

This way the huge else if statement is also not necessary. What do you think?

haerdib: Ofc, the code won't compile. It's only a design suggestion.

teerex/src/tests/tests.rs
```diff
@@ -689,17 +696,253 @@ fn confirm_proposed_sidechain_block_from_shard_neq_mrenclave_errs() {
...
+		let signer7 = get_signer(TEST7_SIGNER_PUB);
+		let shard7 = H256::from_slice(&TEST7_MRENCLAVE);
+
+		//Ensure that enclave is registered
+		assert_ok!(Teerex::register_enclave(
+			Origin::signed(signer7.clone()),
+			TEST7_CERT.to_vec(),
+			URL.to_vec(),
+		));
+		assert_eq!(Teerex::enclave_count(), 1);
```
mullefel, line 730: This seems to be a pattern that is repeated for each test. Can we extract this to a common function?

teerex/src/lib.rs
```diff
@@ -446,6 +500,78 @@ impl<T: Config> Pallet<T> {
 			Err(<Error<T>>::RemoteAttestationTooOld.into())
 		}
 	}
+
+	fn check_block_and_get_latest_header(
```
haerdib, line 504: Clean Code: A function containing an `and` in its name should be split ;)
I'm not even sure if this function is needed.. And if it's needed, the check only which could return a bool or a result.
The function `confirm_sidechain_block` can then be called after a successful check. No hidden function calls and no extra variable needed that don't make sense for "check" only (as the function name suggests)

haerdib (changes requested):

teerex/src/lib.rs
haerdib: The naming of the functions and variables are not yet thought through. Fell free to adapt.

teerex/src/lib.rs
```diff
@@ -446,6 +500,78 @@ impl<T: Config> Pallet<T> {
...
+		while <SidechainBlockHeaderQueue<T>>::contains_key(
+			(shard_id, latest_block_number + 1),
+			latest_block_header.hash(),
+		) && i < early_block_proposal_lenience
+		{
+			let header = <SidechainBlockHeaderQueue<T>>::take(
+				(shard_id, latest_block_number + 1),
+				latest_block_header.hash(),
+			);
+			<SidechainBlockHeaderQueue<T>>::remove_prefix(
```
haerdib, line 556: Why not simple `remove`?

teerex/src/lib.rs
```diff
@@ -446,6 +500,78 @@ impl<T: Config> Pallet<T> {
...
+		shard_id: ShardIdentifier,
+		sender: &T::AccountId,
+		sender_index: u64,
+		mut latest_block_header: Header,
+	) {
+		let mut latest_block_number = latest_block_header.block_number;
+		let early_block_proposal_lenience = T::EarlyBlockProposalLenience::get();
+		let mut i: u64 = 0;
+		while <SidechainBlockHeaderQueue<T>>::contains_key(
+			(shard_id, latest_block_number + 1),
```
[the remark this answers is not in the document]
mullefel: Good catch - there are a couple of unchecked adds in this function: 3x `latest_block_number + 1` and once `i += 1`. The latter should probably be safe, since we only increment until `early_block_proposal_lenience`. But I'm not sure what the best practice is here - maybe all the potential overflow/underflow operations should be checked and therefore also the increment?

teerex/src/lib.rs
mullefel: Agreed that the function name is misleading. You could still have that function, called something with 'add' in it. And then assign it to variables with proper names (just as Bigna proposed). That way, you hide all the `checked_add` complexity (which contrary to Bigna's proposal I don't think belong in this top-level function), but still have readable code.

teerex/src/lib.rs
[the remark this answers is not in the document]
haerdib: Substrate seems to have decided to go for the `saturated_add`(https://github.com/paritytech/substrate/blob/master/frame/referenda/src/lib.rs). There you don't have to handle the error, which might be the way to go for a limited number just as `i`

teerex/src/lib.rs
gaudenzkessler: the plan is to extract this to a sidechain specific pallet (#49). This also resolves the confusion.

primitives/teerex/Cargo.toml
```diff
@@ -24,6 +26,7 @@ default = ["std"]
 std = [
     "codec/std",
     "scale-info/std",
```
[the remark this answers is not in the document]
gaudenzkessler: ok but that costs extra.

primitives/teerex/src/lib.rs
gaudenzkessler: the plan is to extract this to a sidechain specific pallet (https://github.com/integritee-network/pallets/issues/49, in progress). This also resolves the confusion.

mullefel: Even in its own pallet, it should be named `SidechainBlockHeader` IMO - it's not the pallet that creates the ambiguity, but the fact that we're in a parentchain context (which will not change with a different pallet)

gaudenzkessler: ok

gaudenzkessler: But why SidechainBlockHeader and not SidechainHeader? Is it not obvious that it is the header of a block?

teerex/src/lib.rs
gaudenzkessler: more precise? above the pallet definition or inside? here: https://github.com/integritee-network/pallets/blob/9b7528f2d287129dc7e127475b195a3ac91ca060/teerex/src/lib.rs#L41?

teerex/src/lib.rs
gaudenzkessler: I prefer to see 4 logically seperated cases. This is what the design relies on. The checks should done by the same method to prevent chaning one statement could introduce a gap or intersection of cases.

teerex/src/lib.rs
gaudenzkessler: I inlined the function

teerex/src/lib.rs
gaudenzkessler: we want to remove not a specific entry but all entries with the first key https://docs.rs/frame-support/2.0.0-rc4/frame_support/storage/trait.StorageDoubleMap.html#tymethod.remove_prefix

teerex/src/lib.rs
haerdib: Hm. I don't care that much, but seems to be convention to have non-runtime related types outside the pallet. So yes, just like in the teerex pallet.

teerex/src/lib.rs
haerdib: But it's very hard to see which case leads to what (high cognitive complexity, however it's called). And I don't see why an else if statement would ensure that no intersection is happening? You could just change the check condition of one and there you go - intersection is there... it's just a matter of ordering which condition get checked first, but that's the case in my design proposal as well.

But I'll concede on that matter, as you guys have way more experience on that.

clangenb (approved):
Looks good to me now, thanks!

mullefel (approved):
LGTM, we can proceed to the next step. Thanks ☺️

haerdib (approved):
I concede ;)

PR #61, opened 2022-06-10
Move sidechain block and header to pallet repo

gaudenzkessler:
Add sidechain block and header to sidechain pallet.


primitives/sidechain/Cargo.toml
```diff
@@ -8,14 +8,15 @@ license = "Apache-2.0"
...
-codec = { version = "3.0.0", default-features = false, features = ["derive"], package = "parity-scale-codec" }
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive", "full"] }
 scale-info = { version = "2.0.1", default-features = false, features = ["derive"] }
 serde = { version = "1.0.130", default-features = false }
 
 
 # substrate dependencies
 sp-io = { version = "6.0.0", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.20" }
 sp-core = { version = "6.0.0", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.20" }
+sp-runtime = { version = "6.0.0", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.20"}
```
clangenb, line 19: I think you forgot enabling the `std`

primitives/sidechain/src/constants.rs
```diff
@@ -0,0 +1,22 @@
...
+	See the License for the specific language governing permissions and
+	limitations under the License.
+
+*/
+
+//! Sidechain constants
+
+// RPC method names.
+pub const RPC_METHOD_NAME_IMPORT_BLOCKS: &str = "sidechain_importBlock";
+pub const RPC_METHOD_NAME_FETCH_BLOCKS_FROM_PEER: &str = "sidechain_fetchBlocksFromPeer";
```
clangenb, line 22: I think this is really only relevant for the worker, this should not be in here.

I would put it in this crate: https://github.com/integritee-network/worker/tree/master/sidechain/rpc-handler

clangenb (commented):
Two minor comments. It looks good otherwise!

echevrier (approved):
I have some problems with the code, but I think you have copied it from worker and it is not the purpose of this PR to improve it.

PR #62, opened 2022-06-16
[sidechain-primitives] make no_std compatible + add full-crypto feature flag

clangenb:
So we do not have to import the massive crypto primitives into the parachain runtime

### DIFF
--- a/primitives/sidechain/src/traits/mod.rs
+++ b/primitives/sidechain/src/traits/mod.rs
@@ -22,7 +22,7 @@
 
 use codec::{Decode, Encode};
 use core::hash::Hash;
-use sp_core::{blake2_256, Pair, Public, H256};
+use sp_core::{blake2_256, Public, H256};
 use sp_runtime::traits::Member;
 use sp_std::{fmt::Debug, prelude::*};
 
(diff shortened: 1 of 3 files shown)


primitives/sidechain/src/traits/mod.rs
```diff
@@ -21,14 +21,13 @@
 //! some generic structs.
 
 use codec::{Decode, Encode};
-use core::hash::Hash;
-use sp_core::{blake2_256, Pair, Public, H256};
-use sp_runtime::traits::Member;
+use sp_core::{crypto::Public, H256};
```
clangenb, line 24: @gaudenzkessler This one is mean. Public is available with `no_std` but just not re-exported at top level.

primitives/sidechain/src/traits/mod.rs
```diff
@@ -41,7 +40,7 @@ pub trait Header: Encode + Decode + Clone {
 
 	/// get the `blake2_256` hash of the header.
 	fn hash(&self) -> H256 {
-		self.using_encoded(blake2_256).into()
+		self.using_encoded(BlakeTwo256::hash).into()
```
clangenb, line 43: Same function, but only `BlakeTwo256` is available in no_std, without activating the `full_crypto` feature.

primitives/sidechain/src/traits/mod.rs
```diff
@@ -138,28 +137,37 @@ pub trait SignedBlock: Encode + Decode + Send + Sync + Debug + Clone {
...
-{
-	fn sign_block<P: Pair>(self, signer: &P) -> SignedSidechainBlock
-	where
-		<SignedSidechainBlock as SignedBlock>::Signature: From<<P as sp_core::Pair>::Signature>;
-}
+#[cfg(feature = "full_crypto")]
+pub use crypto::*;
+
+#[cfg(feature = "full_crypto")]
+mod crypto {
```
clangenb, line 144: Signing logic is only compiled with `full-crypto`. We do not need it in the runtime, and we don't want all of these dependencies there.

Niederb (approved):
LGTM

PR #74, opened 2022-07-22
Ignore CLion files

masapr:


clangenb (changes requested):
This is where I personally disagree with the rust guidelines. Having a `rust-toolchain.toml` together with a `Cargo.lock` gives us deterministic builds. 
Also
* With git dependencies you see exactly, which commit has been used in the Cargo.lock file, and we have plenty of git deps.
* When we update the git deps, and substrate in particular, you can see in the Cargo.lock-diff if something has gone wrong, like that we are pulling in two different substrate versions.

masapr:
But it wasn't checked in before. Should I simply check in mine?

clangenb:
Ah what? It used to be checked in, I am definitely in favour of checking it in. Maybe we can discuss this in the daily afterwards. :)

masapr:
changed to only ignore CLion files. Not Cargo.lock

clangenb (approved):
LGTM!

PR #75, opened 2022-08-01
[types] remove api-client definitions and dependency

### LINKED ISSUE #72: remove dependency on the api-client
The refactoring of the primtives has brought the dependency of the api-client here. In my opinion, this brings unnecessary dependency complexity into the pallets and should be avoided because it also makes the upgrade process more tedious, which was at least for the pallets fairly easy until now as they were decoupled from our other repositories.

clangenb:
* Closes #72.

* https://github.com/integritee-network/worker/pull/892 the worker re-introduces them here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,55 +13,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ac-compose-macros"
-version = "0.1.0"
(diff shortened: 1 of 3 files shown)


haerdib (approved):
Great, thanks !

Issue #76, opened 2022-08-15
Implement sidechain finality - simplified version


mullefel:
As part of #855, we need to change our current implementation of the sidechain finality in the sidechain pallet.

* Instead of storing a list of finalized sidechain headers, we only store the latest finalized sidechain block (identified by block number and hash)
* In this simplified first version, we don't employ a majority algorithm to determine if a block is considered finalized. Instead, just finalize everything we receive from the first validateer. First being the validateer registered first at pallet Teerex.

- [x] create a new pallet call new `confirm_imported_sidechain_block` as a copy from 'confirm_proposed_sidechain_block'
- [x] make all tests use the new call
- [x] refactor 'confirm_imported_sidechain_block' to use a simple storage value, with blocknumber and hash.
- [x] change 'confirm_imported_sidechain_block', so that it receives blocknumber and hash (not the full SidechainHeader) and adjust the tests accordingly (if needed).

- [x] implement [#922](https://github.com/integritee-network/worker/issues/922) (i. e. use the new pallet-call, don't use the old one anymore).

- [x] remove pallet call 'confirm_proposed_sidechain_block'
- [x] Remove dependency on `sidechain-primitives` crate (since we don't need `SidechainHeader` anymore, we only store hash and number) https://github.com/integritee-network/worker/issues/924
- [x] Increase transaction_version in referencing integritee-node (see comment below).

masapr:
@mullefel Can you check, if you're ok with my changes? ... sorry for being picky with the wording here, but this task is not a refactoring, we change the logic (because we only process calls from the "first" validateer).

mullefel:
Of course, yes! You're right, it's not a refactoring but a change in the whole logic 👍

masapr:
Input from @clangenb : 
 if we remove the old confirm sidechain block in the future. This requires a bump in the transaction version in the runtime. From the docs:

```
/// All existing dispatches are fully compatible when this number doesn't change. If this
/// number changes, then `spec_version` must change, also.
///
/// This number must change when an existing dispatchable (module ID, dispatch ID) is changed,
/// either through an alteration in its user-level semantics, a parameter
/// added/removed/changed, a dispatchable being removed, a module being removed, or a
/// dispatchable/module changing its index.
///
/// It need *not* change when a new module is added or when a dispatchable is added.
pub transaction_version: u32,
```

PR #78, opened 2022-08-18
Sp/simple sc finality

masapr:
Implementation of #76

After the worker uses the new call here, the cleanup of the old code still needs to be done.

### DIFF
--- a/sidechain/src/lib.rs
+++ b/sidechain/src/lib.rs
@@ -199,7 +199,6 @@ pub mod pallet {
 			}
 
 			let lenience = T::EarlyBlockProposalLenience::get();
-			let mut latest_header = <LatestSidechainHeader<T>>::get(shard_id);
 			let mut latest_confirmation = <LatestSidechainBlockConfirmation<T>>::get(shard_id);
 			let latest_block_number = latest_confirmation.block_number;
 			let block_number = confirmation.block_number;
@@ -210,32 +209,19 @@ pub mod pallet {
 			} else if block_number > Self::add_to_block_number(latest_block_number, 1)? {
 				// Block is too early and stored in the queue for later import.
 				if !<SidechainBlockConfirmationQueue<T>>::contains_key((shard_id, block_number)) {
-					<SidechainHeaderQueue<T>>::insert(
-						(shard_id, block_number),
-						header.parent_hash,
-						header,
-					);
 					<SidechainBlockConfirmationQueue<T>>::insert(
 						(shard_id, block_number),
 						confirmation,
 					);
 				}
 			} else if block_number == Self::add_to_block_number(latest_block_number, 1)? {
-				Self::confirm_sidechain_block(
-					shard_id,
-					header,
-					confirmation,
-					&sender,
-					sender_index,
-				);
-				latest_header = header;
+				Self::confirm_sidechain_block(shard_id, confirmation, &sender, sender_index);
 				latest_confirmation = confirmation;
 
 				Self::finalize_blocks_from_queue(
 					shard_id,
 					&sender,
 					sender_index,
-					latest_header,
 					latest_confirmation,
 				)?;
 			} else {
@@ -259,7 +245,6 @@ pub mod pallet {
 impl<T: Config> Pallet<T> {
 	fn confirm_sidechain_block(
 		shard_id: ShardIdentifier,
-		header: SidechainHeader,
 		confirmation: SidechainBlockConfirmation,
 		sender: &T::AccountId,
 		sender_index: u64,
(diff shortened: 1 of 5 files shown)


sidechain/src/tests.rs
```diff
@@ -26,259 +26,220 @@ fn get_signer(pubkey: &[u8; 32]) -> AccountId {
...
-			Sidechain::latest_sidechain_header(shard7).block_data_hash,
-			header3a.block_data_hash
-		);
+		register_enclave7();
+
+		let hash = H256::random();
+
+		assert_ok!(confirm_block7(1, H256::random(), true));
+		assert_eq!(Sidechain::latest_sidechain_block_confirmation(shard7).block_number, 1);
+		assert_ok!(confirm_block7(3, hash, false));
```
haerdib, line 106: I know this hasn't been here before, but it was a little pain to understand this test. Can you add a comment here like
`// Queue block number 3. Should not be imported right now, because confirmation of block number 2 is missing`

sidechain/src/tests.rs
```diff
@@ -26,259 +26,220 @@ fn get_signer(pubkey: &[u8; 32]) -> AccountId {
...
+		assert_eq!(Sidechain::latest_sidechain_block_confirmation(shard7).block_number, 1);
+		assert_ok!(confirm_block7(3, H256::random(), false));
```
haerdib, line 108: `//  Queue another block with number 3. Should not be imported at all, because it was not the first one`

sidechain/src/tests.rs
```diff
@@ -26,259 +26,220 @@ fn get_signer(pubkey: &[u8; 32]) -> AccountId {
...
+		assert_eq!(Sidechain::latest_sidechain_block_confirmation(shard7).block_number, 1);
+		assert_ok!(confirm_block7(2, H256::random(), true));
```
haerdib, line 110: `// Queue block number 2. Block number 3 should now be confirmed as well.`

haerdib: Somthing like that? :)

haerdib (commented):
Looks good! Some small comments only

sidechain/src/tests.rs
```diff
@@ -26,259 +26,220 @@ fn get_signer(pubkey: &[u8; 32]) -> AccountId {
...
 			Origin::signed(signer7.clone()),
 			shard7.clone(),
-			header.clone(),
+			1,
+			hash
 		));
 
 		let expected_event =
-			Event::Sidechain(SidechainEvent::ProposedSidechainBlock(signer7, block_hash));
+			Event::Sidechain(SidechainEvent::ImportedSidechainBlock(signer7, hash));
```
clangenb, line 46: Hmm, I think I would pick another name for the event, as we don't import any blocks here.

What would you think about:
* SidechainBlockFinalized/FinalizedSidechainBlock

Sounds a bit more epic and it is actually what is happening.

clangenb (commented):
I have one remark about the event name, but it looks good otherwise!

sidechain/src/lib.rs
```diff
@@ -45,6 +48,14 @@ pub mod pallet {
 	#[pallet::without_storage_info]
 	pub struct Pallet<T>(PhantomData<T>);
 
+	#[derive(PartialEq, Eq, Clone, Encode, Decode, Debug, Copy, Default, TypeInfo)]
+	#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+	pub struct SidechainBlockConfirmation {
+		pub block_number: u64,
+		/// Hash of the block header. TODO check
```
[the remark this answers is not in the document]
masapr: Not TODO anymore, but should we stick to the name "block_header_hash"? Or change to "hash" or "block_hash"? The plan is now to really use the hash of the block-header, so the current name would generally be correct.

sidechain/src/tests.rs
masapr: Thanks for the input, I also adjusted one of the function names accordingly. Fits much better now :-)

sidechain/src/lib.rs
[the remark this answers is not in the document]
haerdib: Depends. Is the future plan to remove block hash altogether, or stick to the two different headers (e.g. block header hash & block hash)?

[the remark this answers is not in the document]
clangenb: I have not 100% wrapped my had around the hashing either. I will think about hit.

clangenb (approved):
Nice, looks good to me now!

masapr:
I did a rebase and removed unused code. I think it can be merged after that.

PR #79, opened 2022-08-26
Update Cargo.toml files

Niederb:
- Switch to edition 2021
- Standardize the fields license, authors, homepage, repository
- I would update the remaining repos in a similar way. 
- Note that there is the pallet-claims in this repo from parity which I did not change


clangenb (approved):
Nice, looks good to me!

PR #81, opened 2022-09-01
rename parameter

masapr:


clangenb:
I will close this PR as it is outdated, we have changed the finality logic since then. :)

Issue #56, opened 2022-05-16
Remove outdated `TODO` comment


haerdib:
I believe the `TODO` comment
https://github.com/integritee-network/pallets/blob/4d1b0e9d412b4e3b94f3e5afd853213b6ecc97b7/teerex/src/lib.rs#L175-L177

is outdated now?
We already have https://github.com/integritee-network/pallets/blob/4d1b0e9d412b4e3b94f3e5afd853213b6ecc97b7/teerex/src/lib.rs#L418-L422 implemented.

clangenb:
Yes, this can be removed.

PR #82, opened 2022-09-06
Remove outdated todo comment

Niederb:
Extremely proud of this achievement 😄

### DIFF
--- a/teerex/src/lib.rs
+++ b/teerex/src/lib.rs
@@ -172,9 +172,6 @@ pub mod pallet {
 			Ok(().into())
 		}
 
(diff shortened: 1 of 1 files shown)


clangenb (commented):
Ehhhm, I am very proud of you. However, why is this comment obsolete? :)

haerdib:
https://github.com/integritee-network/pallets/issues/56 :P

clangenb:
But IMO the function needs to be removed to.

Niederb:
I think it still makes sense to allow to unregister the enclave (for example we could do it during worker shutdown).

clangenb:
Ok, I agreed 👍

PR #86, opened 2022-09-13
remove weights of deprecated call

gaudenzkessler:
remove weights of  confirm_proposed_sidechain_block since call was removed in #84


haerdib (approved):
Ah, so that was the problem. Thanks for looking into it!

PR #92, opened 2022-09-20
Add block number to ProcessedParentchainBlock event

gaudenzkessler:
[Belongs to 838](https://github.com/integritee-network/worker/issues/838)

### DIFF
--- a/teerex/src/lib.rs
+++ b/teerex/src/lib.rs
@@ -76,7 +76,7 @@ pub mod pallet {
 		Forwarded(ShardIdentifier),
 		ShieldFunds(Vec<u8>),
 		UnshieldedFunds(T::AccountId),
-		ProcessedParentchainBlock(T::AccountId, H256, H256),
+		ProcessedParentchainBlock(T::AccountId, H256, H256, T::BlockNumber),
 	}
 
 	// Watch out: we start indexing with 1 instead of zero in order to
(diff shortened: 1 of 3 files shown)


clangenb (approved):
LGTM

teerex/src/lib.rs
```diff
@@ -76,7 +76,7 @@ pub mod pallet {
 		Forwarded(ShardIdentifier),
 		ShieldFunds(Vec<u8>),
 		UnshieldedFunds(T::AccountId),
-		ProcessedParentchainBlock(T::AccountId, H256, H256),
+		ProcessedParentchainBlock(T::AccountId, H256, H256, u64),
```
haerdib, line 79: Is it not possible to use the runtime Blocknumber? E.g. T::BlockNumber?

clangenb: Should be possible, but this would require a change in the extrinsic signature too, so I guess, we leave it as it is.

haerdib: Would it? Either way: Use u32 instead of u64. https://github.com/integritee-network/parachain/blob/master/polkadot-parachains/integritee-runtime/src/lib.rs#L751-L752

haerdib:
Could you change the PR to something more descriptive?

teerex/src/lib.rs
clangenb: @gaudenzkessler told me that the signature change is required anyhow, so I think we should change it.

PR #93, opened 2022-09-21
Propose every nth block as finality candidate instead of all the blocks

gaudenzkessler:
belongs to [#1003](https://github.com/integritee-network/worker/issues/1003)

### DIFF
--- a/sidechain/src/lib.rs
+++ b/sidechain/src/lib.rs
@@ -79,6 +79,10 @@ pub mod pallet {
 	pub type SidechainBlockConfirmationQueue<T: Config> =
 		StorageMap<_, Blake2_128Concat, ShardBlockNumber, SidechainBlockConfirmation, ValueQuery>;
 
+	#[pallet::storage]
+	pub type SidechainBlockFinalizationCandidates<T: Config> =
+		StorageMap<_, Blake2_128Concat, ShardIdentifier, Vec<u64>, ValueQuery>;
+
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
 		/// The integritee worker calls this function for every imported sidechain_block.
@@ -87,7 +91,7 @@ pub mod pallet {
 			origin: OriginFor<T>,
 			shard_id: ShardIdentifier,
 			block_number: u64,
-			block_number_diff: u64,
+			next_finalization_candidate_block_number: u64,
 			block_header_hash: H256,
 		) -> DispatchResultWithPostInfo {
 			let confirmation = SidechainBlockConfirmation { block_number, block_header_hash };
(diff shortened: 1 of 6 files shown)


sidechain/src/lib.rs
```diff
@@ -118,15 +119,19 @@ pub mod pallet {
 			if block_number > Self::add_to_block_number(latest_block_number, lenience)? {
 				// Block is far too early and hence refused.
 				return Err(<Error<T>>::BlockNumberTooHigh.into())
-			} else if block_number > Self::add_to_block_number(latest_block_number, 1)? {
+			} else if block_number >
```
haerdib, line 122: Since the validateers choose the finality candidates, and that is one per parentchain block, I don't think we need this lenience and block queue any more, do we? @clangenb

haerdib: Emphasizing this again - do we still need this complicated logic in the pallet? I believe we can remove the block queue fully.

sidechain/src/lib.rs
```diff
@@ -110,47 +108,41 @@ pub mod pallet {
...
-				)?;
+				Ok(().into())
+			} else if block_number > latest_block_number {
+				// Block does not belong to the finalized ones.
+				Err(<Error<T>>::ReceivedUnexpectedSidechainBlock.into())
 			} else {
 				// Block is too late and hence refused.
-				return Err(<Error<T>>::OutdatedBlockNumber.into())
+				Err(<Error<T>>::OutdatedBlockNumber.into())
 			}
```
clangenb, line 134: I think the branches here are a bit confusing now. I guess we simplify the code here to:

```rust

if block_number == finalization_candidate_block_number {
    Self::finalize_block(...)
    /// see comment above
   <SidechainBlockFinalizationCandidate<T>>::insert(
				shard_id,
				next_finalization_candidate_block_number,
   );
} else {
    Err(<Error<T>>::ReceivedUnexpectedSidechainBlock.into())
}
```

sidechain/src/tests.rs
```diff
@@ -39,6 +39,7 @@ fn confirm_imported_sidechain_block_works_for_correct_shard() {
 			Origin::signed(signer7.clone()),
```
clangenb, line 39: I would like to see some asserts about the `SidechainBlockFinalizationCandidate` here.

clangenb (commented):
Looks good in general. I have some minor remarks though. :)

OverOrion:
LGTM, thank you, I especially liked that the function `finalize_blocks_from_queue` has been deleted. :rocket:

sidechain/src/lib.rs
coax1d: This works even better or more conventional would be to use ensure.

```rust
// Check error condition
ensure!(block_number == finalization_candidate_block_number, Error::<T>::ReceivedUnexpectedSidechainBlock);
// Now you are in your valid state now mutate storage.
Ok(().into())
```

coax1d (commented):
LGTM

sidechain/src/lib.rs
clangenb: nice, i agree.

sidechain/src/lib.rs
```diff
@@ -110,47 +108,41 @@ pub mod pallet {
...
+			let latest_confirmation = <LatestSidechainBlockConfirmation<T>>::get(shard_id);
 			let latest_block_number = latest_confirmation.block_number;
 			let block_number = confirmation.block_number;
+			let finalization_candidate_block_number =
+				<SidechainBlockFinalizationCandidate<T>>::try_get(shard_id).unwrap_or(1);
+
+			<SidechainBlockFinalizationCandidate<T>>::insert(
+				shard_id,
+				next_finalization_candidate_block_number,
+			);
```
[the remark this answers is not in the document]
clangenb: The verification is simply that we have the correct sidechain block, aka that the block number equals the expected one. So it just needs to be after the ensure you introduced.

clangenb (commented):
Nice, thanks for adding the asserts in the tests! Only one thing remains. :)

PR #94, opened 2022-09-26
Add cargo.toml formatting with taplo

echevrier:
- best config to match coding guidelines
- according to https://github.com/integritee-network/worker/issues/957
- enforces formatting of Cargo.toml in the CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,3 +66,15 @@ jobs:
         with:
           name: license
           path: license.html
+  fmt:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install taplo
+        run: cargo install taplo-cli --locked
+      - name: Cargo.toml fmt
+        run: taplo fmt --check
+      - name: Fail-fast; cancel other jobs
+        if: failure()
+        uses: andymckay/cancel-action@0.2
+
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,8 +77,10 @@ jobs:
     container: "tamasfe/taplo:0.7.0-alpine"
     steps:
       - uses: actions/checkout@v3
+
       - name: Run Taplo fmt
         run: taplo fmt --check
+
       - name: Fail-fast; cancel other jobs
         if: failure()
         uses: andymckay/cancel-action@0.2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
 #                 cargo clippy -- -D warnings Todo: Fix #10
         ]
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       # With rustup's nice new toml format, we just need to run rustup show to install the toolchain
       # https://github.com/actions-rs/toolchain/issues/126#issuecomment-782989659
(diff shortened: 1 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -51,3 +51,16 @@ jobs:
 
       - name: ${{ matrix.check }}
         run: ${{ matrix.check }}
+
+  fmt:
```
haerdib, line 55: I belive it would make sense to put this as an extra step in the previous job instead of a separate one. It depends on the rust toolchain as well as on the environment.

E.g. put the install taplo as an additional step after the rust toolchain installation and add the taplo fmt --check to the matrix.

haerdib (commented):
For me the PR title is a little misleading. It's not a `cargo fmt` but a `taplo fmt` . What do you think?

.github/workflows/ci.yml
echevrier: This would install taplo for each check.

echevrier:
> For me the PR title is a little misleading. It's not a `cargo fmt` but a `taplo fmt` . What do you think?
Add cargo.toml formatting with taplo ?

.github/workflows/ci.yml
haerdib: No, that shouldn't be the case if you put the taplo install as a separate step, just like the rust-toolchain:

```
      - name: Install Taplo
        run: cargo intall ...
```

echevrier: Do you mean like here https://github.com/scs/substrate-api-client/actions/runs/3136637660/workflow#L53 ?
Because this workflow results in installing taplo each check : https://github.com/scs/substrate-api-client/actions/runs/3136637660/jobs/5093805494

haerdib: Ah, interesting. I wasn't aware of that. No, that's not what we want. Still, it should be done after the rustup show, so I think doing it like @Niederb did here: https://github.com/integritee-network/pallets/pull/98/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f makes sense.

Niederb: They way I did it it gets executed multiple times, but the Github action I use has some caching so the installation is only 2 seconds if it is in the cache. I thought of doing it similar to @echevrier but I believed I would need to install rust on another machine... but looking at the implementation here I'm not sure anymore. Maybe this is the better approach. I still think it makes sense to use the Github action in order to have the caching.

haerdib: I see. Then I start to believe the matrix itself is the wrong approach for this.  If all steps are repeated, even though they don't need to be.. such as checkout, rust installation and now the licence check. 

I'd recommend removing the matrix then altogether and add them as simple steps instead. @clangenb do you agree?

Niederb: I also think that the matrix approach uses quite a lot of resources for a speedup that is probably very small.

mullefel: I think it's fine to run it in sequence (as is proposed now?). Installing taplo for each run IMO does not matter much, since the other jobs that run in parallel, take much longer anyway.

mullefel: As alternative, you could use the docker image of taplo: `tamasfe/taplo`. GitHub Actions allows you to run a job directly based on a docker image - that way you wouldn't have to install taplo, just pull the image.

mullefel (approved):
LGTM - just as we discussed! Looking at the timings of the last run, the Taplo job is now really fast, 10 s. The license check takes more than 1 min, that proves us right to have it in a separate (concurrent) job.

haerdib (approved):

PR #98, opened 2022-09-28
Add license check to CI

Niederb:


about.hbs
```diff
@@ -0,0 +1,70 @@
...
+            white-space: pre-wrap;
+        }
+    </style>
+</head>
+
+<body>
+    <main class="container">
+        <div class="intro">
+            <h1>Third Party Licenses</h1>
+            <p>This page lists the licenses of the projects used in cargo-about.</p>
```
clangenb, line 42: Here we should probably change 'cargo-about' to 'Integritee Networks Pallets' repository.

clangenb (commented):
Looks good, cool. I just have one remark!

about.hbs
Niederb: Yeah, probably... 🤣

clangenb (approved):
LGTM

PR #101, opened 2022-09-30
Fix clippy and enable it in CI

### LINKED ISSUE #10: Fix clippy and enable it in CI.

OverOrion:
This PR aims to fix clippy lint warnings and enable it in the CI.

There are two workarounds:
1. `derive_partial_eq_without_eq` workaround - this a known bug: https://github.com/rust-lang/rust-clippy/issues/9413
2. `clippy::type_complexity` - Fixing this with a `type alias` or a `struct` would result in unnecessary extra trait bounds in many places, which is not worth it.

Fixes #10


haerdib (approved):
As always - clippy is right. That were a lot of clippy warnings, thanks so much!

claims/src/lib.rs
```diff
@@ -362,16 +359,25 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			T::MoveClaimOrigin::try_origin(origin).map(|_| ()).or_else(ensure_root)?;
 
-			Claims::<T>::take(&old).map(|c| Claims::<T>::insert(&new, c));
-			Vesting::<T>::take(&old).map(|c| Vesting::<T>::insert(&new, c));
-			Signing::<T>::take(&old).map(|c| Signing::<T>::insert(&new, c));
-			maybe_preclaim.map(|preclaim| {
+			if let Some(c) = Claims::<T>::take(&old) {
+				Claims::<T>::insert(&new, c);
+			}
```
clangenb, line 364: This one makes me curious, do you know why clippy suggests the more verbose versions?

claims/src/lib.rs
```diff
@@ -567,14 +585,13 @@ where
...
-				let signer = Preclaims::<T>::get(who)
-					.ok_or(InvalidTransaction::Custom(ValidityError::SignerHasNoClaim.into()))?;
-				if let Some(s) = Signing::<T>::get(signer) {
-					let e = InvalidTransaction::Custom(ValidityError::InvalidStatement.into());
-					ensure!(&attested_statement[..] == s.to_text(), e);
-				}
+		if let Some(Call::attest { statement: attested_statement }) = call.is_sub_type() {
+			let signer = Preclaims::<T>::get(who).ok_or_else(|| {
+				InvalidTransaction::Custom(ValidityError::SignerHasNoClaim.into())
+			})?;
```
clangenb, line 591: Do we want the closure variant here because it is likely to not get executed and so we safe computational effort on average?

clangenb (commented):
Nice, this has been on hold for 1.5 years :)

PR #102, opened 2022-10-03
Speedup licence check

Niederb:
The license check currently takes a long time because the `cargo` command triggers a download of the nightly toolchain (which is irrelevant for the license check). We can avoid that by explicitly using stable rust to run the license check


clangenb:
We indeed save a few minutes compared to master because the cargo install does not trigger a download of the rust-toolchain any more. Hence, I say that we merge this PR 👍

clangenb (approved):

PR #109, opened 2022-10-23
Add support for generic oracle in pallets

### LINKED ISSUE #68: [teeracle] support arbitrary oracle data and strings
Exchange rate data is of course very relevant for a TEEracle, but we should not confine ourselves to this. If other parachains want to benefit from our TEEracle, the chances are high that we should support arbitrary data. Hence, I suggest we add the following storage:

```
	/// Arbitrary Web2 oracle data
	#[pallet::storage]
	#[pallet::getter(fn exchange_rate)]
	pub(super) type Web2OracleData<T: Config> = StorageDoubleMap<
		_,
		Blake2_128Concat,
		OracleDataName, // Name of the oracle data
		Blake2_128Concat,
		Web2SourceAddress, // The Web2 source that feeds the oracle
		OracleDataBlob,  // Arbitrary encoded Web2 data, we can't and don't need to interpret this. The chain interested in this data knows how to do this.
		ValueQuery,
	>;
```

coax1d:
closes #68

This is the beginning of the refactor for supporting generic oracles.
- Added New Storage Item
- Added New Extrinsic
- Added Tests
- Added Benchmarks
- Added Weights
- Added some substrate related factoring

After this new demo feature is fully complete we should remove the old `update_exchange_extrinsic`. For now seeking backwards compatibility with the current oracle demo.


teeracle/src/mock.rs
```diff
@@ -49,11 +49,11 @@ frame_support::construct_runtime!(
...
-		System: frame_system::{Pallet, Call, Config, Storage, Event<T>},
-		Balances: pallet_balances::{Pallet, Call, Storage, Config<T>, Event<T>},
-		Timestamp: timestamp::{Pallet, Call, Storage, Inherent},
-		Teerex: pallet_teerex::{Pallet, Call, Storage, Event<T>},
-		Exchange: pallet_exchange::{Pallet, Call, Storage, Event<T>},
+		System: frame_system,
+		Balances: pallet_balances,
+		Timestamp: timestamp,
+		Teerex: pallet_teerex,
+		Teeracle: pallet_teeracle,
```
clangenb, line 56: Nice. Out of personal interest. Can you elaborate on this?

teeracle/src/lib.rs
```diff
@@ -151,10 +171,42 @@ pub mod pallet {
...
+		) -> DispatchResultWithPostInfo {
+			let signer = ensure_signed(origin)?;
+			<pallet_teerex::Pallet<T>>::is_registered_enclave(&signer)?;
+			let signer_index = <pallet_teerex::Pallet<T>>::enclave_index(signer);
+			let signer_enclave = <pallet_teerex::Pallet<T>>::enclave(signer_index)
+				.ok_or(pallet_teerex::Error::<T>::EmptyEnclaveRegistry)?;
+
+			ensure!(
+				Self::is_whitelisted(&data_source, signer_enclave.mr_enclave),
+				<Error<T>>::ReleaseNotWhitelisted
```
clangenb, line 189: I wonder about the error's name here, but I guess this was like this before?

clangenb (commented):
Nice. Looks very straightforward and clean! I only have minor questions.

clangenb:
Yes I agree we can remove `update_exchange_extrinsic` in the future, but it is good to keep it for now.

Also, we should remember that this needs a `transaction_version` increase if I am not mistaken?

teeracle/src/mock.rs
coax1d: Before the `construct_runtime!` macro needed you to specify the types you wanted to build for each pallet such as `Call`, `Event` etc. 

However now that is unneeded. The macro can scan and see which types to build from the pallet definition itself. The most you might want is just to number each pallet i.e. `System: frame_system = 1,` etc.

teeracle/src/lib.rs
coax1d: Yea this is me just using the same convention which was in place before to keep consistency.

coax1d:
@clangenb What else would be necessary to see this get pushed through. Should I structure a pr on `integritee-network/parachain` which updates the runtime configuration for this pallet as well bumps the `transaction_version`?

teeracle/src/lib.rs
```diff
@@ -163,6 +215,7 @@ pub mod pallet {
 			let sender_index = <pallet_teerex::Pallet<T>>::enclave_index(sender);
 			let sender_enclave = <pallet_teerex::Pallet<T>>::enclave(sender_index)
 				.ok_or(pallet_teerex::Error::<T>::EmptyEnclaveRegistry)?;
+			// Todo: Never checks data source len
```
OverOrion, line 218: [Really just optional]
```suggestion
			// TODO: Never checks data source len
```

OverOrion (approved):
LGTM, I like how it is much more generic than before!

PR #121, opened 2022-12-07
Pr polkadot v0.9.32

BillyWooo:
upgrade to substrate: polkadot-v0.9.32
- Only bump up to version v0.9.32
- 'Origin' -> 'RuntimeOrigin', etc. renaming

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,9 @@ jobs:
           version: "0.5"
 
       - name: Run license check
-        run: cargo about generate about.hbs > license.html
+        # Explicitly use stable because otherwise cargo will trigger a download of
+        # the nightly version specified in rust-toolchain.toml
+        run: cargo +stable about generate about.hbs > license.html
 
       - name: Archive license file
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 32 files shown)


clangenb:
Hmm, it seems that parity has included a new dependency that is plain GPL-3, how do you suggest we handle this @Niederb?

Can we still be Apache-2? If not, parity can't either, I suppose - haha.

Niederb:
@haerdib You are quick 😄 
@clangenb Seems that this one is already reported but yeah if our main dependency is not strict with the licensing then there will always be issues popping up. 
Our options are quite limited:
- Reporting it upstream or making PRs ourselves
- Changing our licensing

clangenb:
Patching does not work, you can only patch the same version. As this will be fixed in a future release, I simply suggest to accept the licence violation for now.

haerdib:
There's an even simpler solution, I believe: @BillyWooo can you try to run 
`cargo update -p array-bytes` ? That should update it to 4.2, which already includes Apache 2.0 license.

Tested it on the substrate-api-client.

clangenb (approved):
Nice, looks good to me!

PR #122, opened 2022-12-13
Xcm Transactor for Swapping lease slots

coax1d:
Extendable Xcm Transactor which initially supports sending a UMP message to the Relay chain to call 'Registrar::swap'. This is necessary for swapping lease slots between the shell-runtime and Integritee runtime on Kusama.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,7 +33,7 @@ version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
 dependencies = [
- "getrandom 0.2.7",
+ "getrandom 0.2.8",
  "once_cell",
  "version_check",
 ]
@@ -67,9 +67,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.65"
+version = "1.0.66"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "98161a4e3e2184da77bb14f02184cdd111e83bbbcc9979dfee3c44b9a85f5602"
+checksum = "216261ddc8289130e551ddcd5ce8a064710c0d064a4d2895c67151c92b5443f6"
 
 [[package]]
 name = "approx"
@@ -81,19 +81,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "array-bytes"
+version = "4.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f52f63c5c1316a16a4b35eaac8b76a98248961a533f061684cb2a7cb0eafb6c6"
+
+[[package]]
 name = "arrayref"
 version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a4c527152e37cf757a3f78aae5a06fbeefdb07ccc535c980a3208ee3060dd544"
-
-[[package]]
-name = "arrayvec"
-version = "0.4.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cd9fd44efafa8690358b7408d253adf110036b88f55672a933f01d616ad9b1b9"
-dependencies = [
- "nodrop",
-]
 
 [[package]]
 name = "arrayvec"
@@ -109,9 +106,9 @@ checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
 name = "async-trait"
-version = "0.1.57"
+version = "0.1.58"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "76464446b8bc32758d7e88ee1a804d9914cd9b1cb264c029899680b0be29826f"
+checksum = "1e805d94e6b5001b651426cf4cd446b1ab5f319d27bab5c644f61de0a804360c"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -164,9 +161,15 @@ checksum = "6107fe1be6682a68940da878d9e9f5e90ca5745b3dec9fd1bb393c8777d4f581"
 
 [[package]]
 name = "base64"
-version = "0.13.0"
+version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
+checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
+
+[[package]]
+name = "base64ct"
+version = "1.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b645a089122eccb6111b4f81cbc1a49f5900ac4666bb93ac027feaecf15607bf"
 
 [[package]]
 name = "bitflags"
@@ -192,17 +195,7 @@ version = "0.10.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b9cf849ee05b2ee5fba5e36f97ff8ec2533916700fc0758d40d92136a42f3388"
 dependencies = [
- "digest 0.10.3",
-]
-
-[[package]]
-name = "blake2-rfc"
-version = "0.2.18"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5d6d530bdd2d52966a6d03b7a964add7ae1a288d25214066fd4b600f0f796400"
-dependencies = [
- "arrayvec 0.4.12",
- "constant_time_eq",
+ "digest 0.10.5",
 ]
 
 [[package]]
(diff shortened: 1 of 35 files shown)


Cargo.lock
```diff
@@ -2505,12 +3380,36 @@ dependencies = [
...
+ "sp-io 6.0.0",
+ "sp-std 4.0.0",
+ "sp-weights 4.0.0 (git+https://github.com/paritytech/substrate?branch=polkadot-v0.9.32)",
+]
+
+[[package]]
+name = "sp-runtime"
+version = "7.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0edfc5c54c2b31d2f0cf904d472a0bff7125c0c2a2e2330507842e56f9a27444"
```
clangenb, line 3395: We seem to have duplicate dependencies here. So substrate dependencies come from crates.io and some from github, we should try to get rid of the duplicates.

primitives/xcm-transactor/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+	/// - Corresponding Xcm Message for Transacting on this RelayCall
+	///
+	fn construct_transact_xcm(call: Self::RelayCall, weight: XcmWeight) -> Xcm<()>;
+}
+
+#[derive(Encode, Decode, RuntimeDebug)]
+pub enum RegistrarCall {
+	/// Corresponds to the swap extrinsic index within the Registrar Pallet
+	#[codec(index = 3)]
+	Swap(ParaId, ParaId),
```
clangenb, line 52: I prefer named fields when it comes to such enum args, otherwise it is very unclear, e.g.:

```rust
pub enum RegistrarCall {

    Swap {
       this: ParaId
       other: ParaId 
    }
}
```

primitives/xcm-transactor/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+	}
+}
+
+#[cfg(feature = "kusama")]
+pub use ksm::*;
+
+#[cfg(feature = "polkadot")]
+pub use dot::*;
+
+pub struct RelayCallBuilderType;
```
clangenb, line 83: What would you think about droping the `Type` in the `RealyCallBuilderType` name?

primitives/xcm-transactor/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+
+# cumulus
+cumulus-primitives-core = { git = "https://github.com/paritytech/cumulus", branch = "polkadot-v0.9.32", default-features = false }
+
+[features]
+default = ["std"]
+kusama = []
+polkadot = []
+std = [
+    "common-primitives/std",
```
clangenb, line 30: We are missing the `codec/std` here.

xcm-transactor/Cargo.toml
```diff
@@ -0,0 +1,69 @@
...
+
+[dev-dependencies]
+externalities = { package = "sp-externalities", git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.32" }
+frame-benchmarking = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.32" }
+hex-literal = "0.3.2"
+subxt = "0.25.0"
+sp-keyring = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.32" }
+pallet-balances = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.32" }
+test-utils = { path = "../test-utils" }
+timestamp = { package = "pallet-timestamp", git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.32" }
```
clangenb, line 46: I prefer if you did not rename the packages, then the ordering of the entries does also make more sense. :)

xcm-transactor/src/tests.rs
```diff
@@ -0,0 +1,108 @@
...
+		assert_noop!(
+			XcmTransactor::send_swap_ump(
+				RuntimeOrigin::root(),
+				shell_id.into(),
+				ParaId::from(20000)
+			),
+			Error::<Test>::InvalidSwapIds
+		);
+	})
+}
```
clangenb, line 74: We should also test it the other way around. :)

coax1d:
@clangenb @OverOrion The CI might be having some issues with licensing due to xcm and some other polkadot related procedural macros. Also need to talk a bit about weights on the extrinsic and what is necessary and what is not. lmk what you think so far.

clangenb (commented):
Beautiful, I only have a few minor remarks. :)

primitives/xcm-transactor/src/lib.rs
coax1d: I am more than Ok with it

xcm-transactor/Cargo.toml
coax1d: This was copied over mostly from the teeracle. Which ones have been renamed? like timestamp? Actually just realized we have no need for timestamp here anyways.

xcm-transactor/src/tests.rs
coax1d: true

primitives/xcm-transactor/Cargo.toml
coax1d: Nice catch

Cargo.lock
coax1d: Need to speak with you on this :)

coax1d: What is best approach to go about taking care of this in a methodical way?

primitives/xcm-transactor/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+		#[codec(index = 70)]
+		Registrar(RegistrarCall),
+	}
+}
+
+#[cfg(feature = "kusama")]
+pub use ksm::*;
+
+#[cfg(feature = "polkadot")]
+pub use dot::*;
```
OverOrion, line 81: Could you please also extend the CI with these new feature flags?

OverOrion (commented):
Besides some minor things it looks great to me, thank you laying the foundation for it! :rocket:

primitives/xcm-transactor/src/lib.rs
clangenb: Good point @OverOrion!

clangenb:
> @clangenb @OverOrion The CI might be having some issues with licensing due to xcm and some other polkadot related procedural macros.

Ah I see, yeah the issue is with Polkadot's `(GPL-3.0-only)` licence. Hmm, this is kind of a bummer, to my understanding we need to use the less permissive GPL3 licences for at your XCM crates, do you agree @Niederb?

clangenb:
All in all this seems very complete to me! But good point with the benchmark, I would have missed that! I think we should do a proper benchmarking setup considering that this pallet has come to stay :)

Niederb:
> > @clangenb @OverOrion The CI might be having some issues with licensing due to xcm and some other polkadot related procedural macros.
> 
> Ah I see, yeah the issue is with Polkadot's `(GPL-3.0-only)` licence. Hmm, this is kind of a bummer, to my understanding we need to use the less permissive GPL3 licences for at your XCM crates, do you agree @Niederb?

@clangenb I agree. I think the polkadot repository is under GPL-3 and if we use it we don't have much choice...

Cargo.lock
clangenb: You can first check:

```bash

 cargo tree -i -p sp-io                                                                                                                ok | 16:59:35
error: There are multiple `sp-io` packages in your project, and the specification `sp-io` is ambiguous.
Please re-run this command with `-p <spec>` where `<spec>` is one of the following:
  sp-io@6.0.0
  sp-io@7.0.0
  
  cargo tree -i -p sp-io@7.0.0                                                                                                     101 err | 16:59:44
sp-io v7.0.0
├── sp-application-crypto v7.0.0
│   └── sp-runtime v7.0.0
│       └── subxt v0.25.0
│           [dev-dependencies]
│           └── pallet-xcm-transactor v0.1.0 (/home/clang/code/integritee-pallets/xcm-transactor)
└── sp-runtime v7.0.0 (*)
```

Now we know that the dependency does indeed come from the subxt dependency (sp-io was a random choice not, but I have gotten used to it because it is the crate that introduces the duplicate rust lang items if you have it twice.)

So much for the identification. Resolving this can either be done by looking for a subxt version, which also fetches the substrate dependencies from github and from the `polakdot-v0.9.32` branch to be precise, which is quite the nuisance actually, or we patch the substrate versions, but this is also quite the nuisance (you can only patch it if the dependences have the same version).

I am not sure what the best approach is currently. We might want to just ignore it, as it is only a dev-dependency which will (hopefully) not pollute dependants.

coax1d: yea I saw on subxt repo that there is a polkadot-v0.9.x.x branch but only for one version of polkadot lol. I guess we just ignore since it is a dev dependency but probably should pay close attention to this in the future

coax1d:
@clangenb What should be done about the license check failure?

clangenb:
> @clangenb What should be done about the license check failure?

We have to add it to the about.toml and then we also have to change the licence of your new crate to the same licence.

clangenb (approved):
Nice looks good to me!

PR #130, opened 2022-12-22
Assign explicit indexes to pallet calls

### LINKED ISSUE #129: Assign explicit indexes to the pallet calls.
We now get warnings if we don't assign explicit indexes to the pallet calls for now, the warnings have just been silenced by adding the pallet in the new `dev` mode, but we should handle this soon because it is a low-hanging fruit.

OverOrion:
Fixes #129.


clangenb (approved):
Nice, thanks for the enormously fast fix!

PR #132, opened 2022-12-27
Updates to xcm-transactor-pallet to allow parachain to import successfully

coax1d:
My lesson to learn that the CI for the `pallets` repository does not catch issues when importing said pallet into a the parachain runtime. This fixes the issue of importing the xcm-transactor pallet.


clangenb (approved):
Haha, yes, this is unfortunately true. This would also speak for a monorepo architecture to easily integrate while testing...

Thanks a lot, looks good to me!

PR #133, opened 2022-12-28
Adding custom origin for xcm-transactor pallet

coax1d:
- Need to add custom origin for governance vote to suffice for making the `send_swap_ump()` call


clangenb (approved):
Perfect!

PR #137, opened 2023-01-11
[xcm-transactor] expose xcm args in swap call

clangenb:
We had the issue that our xcm transact failed on the real Kusama chain because the xcm weights are different from the rococo-local setup. Hence, we should expose teh xcm values as args, so we don't run into the issue again that the call fails and there is nothing we can do without a runtime upgrade.

Tested that it works on a rococo local setup

### DIFF
--- a/xcm-transactor/src/lib.rs
+++ b/xcm-transactor/src/lib.rs
@@ -84,7 +84,7 @@ pub mod pallet {
 		///  Rococo-Local as of 11.01.2022:
 		///		* xcm_weight: 10_000_000_000
 		///		* buy_execution_weight: 500_000_000
-		///  Kusama: do be defined, but the weights will be higher than on Rococo-Local
+		///  Kusama: to be defined, but the weights will be higher than on Rococo-Local
 		///
 		#[pallet::call_index(0)]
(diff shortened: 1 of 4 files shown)


xcm-transactor/src/lib.rs
```diff
@@ -75,35 +72,37 @@ pub mod pallet {
 
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
-		// This function should really only be called once via governance
-		// There is chance for it to be called another time or two in the future
-		// This weight is ARBITRARY.
+		/// Send swap instruction to the relay chain to swap the slot lease of our two parachains.
+		/// This needs to be done from within a pallet as the `XCM` origin must be the parachain
+		/// itself.
+		///
```
OverOrion, line 78: [Style only]
I think that one empty line is enough.

OverOrion (approved):
I had a minor stylistic remark only, LGTM, thank you for this!

xcm-transactor/src/lib.rs
clangenb: Haha, I will gladly do this for you!

PR #140, opened 2023-01-26
Detect an underflow if the timestamp is zero

Niederb:
Currently there is an underflow if the timestamp is 0. I'm not quite sure how the timestamp-pallets works, but I think we should at least detect it.

### DIFF
--- a/teerex/src/lib.rs
+++ b/teerex/src/lib.rs
@@ -453,7 +453,6 @@ impl<T: Config> Pallet<T> {
 		if minimum == 0 {
 			log::error!("Invalid time in unregister_silent_workers. Is the timestamp pallet properly configured?")
 		}
-		let minimum = (now - T::MaxSilenceTime::get()).saturated_into::<u64>();
 		let silent_workers = <EnclaveRegistry<T>>::iter()
 			.filter(|e| e.1.timestamp < minimum)
 			.map(|e| e.1.pubkey);
--- a/teerex/src/lib.rs
+++ b/teerex/src/lib.rs
@@ -25,7 +25,10 @@ use frame_support::{
 };
 use frame_system::{self, ensure_signed};
 use sp_core::H256;
-use sp_runtime::traits::{CheckedSub, SaturatedConversion};
+use sp_runtime::{
+	traits::{CheckedSub, SaturatedConversion},
+	Saturating,
+};
 use sp_std::{prelude::*, str};
 use teerex_primitives::*;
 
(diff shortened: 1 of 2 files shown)


teerex/src/lib.rs
```diff
@@ -446,6 +446,13 @@ impl<T: Config> Pallet<T> {
 	}
 
 	fn unregister_silent_workers(now: T::Moment) {
+		let minimum = now
+			.checked_sub(&T::MaxSilenceTime::get())
+			.unwrap_or_default()
+			.saturated_into::<u64>();
+		if minimum == 0 {
+			log::error!("Invalid time in unregister_silent_workers. Is the timestamp pallet properly configured?")
```
Niederb, line 454: The question is what to do here? I think a log message is at least better than ignoring the underflow.

teerex/src/lib.rs
```diff
@@ -446,7 +446,13 @@ impl<T: Config> Pallet<T> {
 	}
 
 	fn unregister_silent_workers(now: T::Moment) {
-		let minimum = (now - T::MaxSilenceTime::get()).saturated_into::<u64>();
+		let minimum = now
+			.checked_sub(&T::MaxSilenceTime::get())
+			.unwrap_or_default()
```
clangenb, line 451: I think there exists a `saturating_sub` which is  `checked_sub().unwrap_or_default()`, but I am not sure.

teerex/src/lib.rs
clangenb: Timestamp has using system time under the hood. However, I am fine which catching this, but maybe we want to early return in this case?

teerex/src/tests/test_cases.rs
```diff
@@ -159,6 +159,21 @@ fn add_and_remove_enclave_works() {
 	})
 }
 
+#[test]
+fn add_enclave_without_timestamp() {
```
clangenb, line 163: maybe we can be more descriptive here: `add_enclave_without_timestamp_fails`

clangenb (commented):
Nice, thanks for the detail analysis again!

teerex/src/lib.rs
OverOrion: Indeed, silently ignoring is way worse.

I am unsure about the impact though. Does it make sense to continue execution in case the timestamp pallet is not configured properly? If not, then I suggest bailing out here.

clangenb: Yeah, i mean this is very unlikely scenario. The timestamp is set by the timestamp inherent. Inherents are verified by all the other nodes when importing a block to be plausible because, another node can only verify it the timestamp inherent of a block is in reasonable range, but not verify it exactly. Hence, having a wrong value here, would mean that a substantial fraction of the blockchain nodes of the network have a wrong system time configured.

OverOrion (approved):
LGTM, thank you!

teerex/src/lib.rs
Niederb: Yeah, the most likely scenario is a unit test (how I found it) or some other small test setup I would say. Should not happen in production.

clangenb (approved):
Thanks a lot! Looks good to me!

PR #147, opened 2023-01-30
Lit polkadot v0.9.37

BillyWooo:
one more Cargo.lock update.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2287,6 +2287,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "pbkdf2"
+version = "0.11.0"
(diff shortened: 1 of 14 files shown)


BillyWooo:
@clangenb Sorry, missed this lock in last PR. 
Could you please also create a new branch: polkadot-v0.9.37 ? Thanks.

BillyWooo:
> Ah yeah. Maybe we should revisit, why the fuzzer recommend to have a separate workspace. I don't like that we have two locks here now.

The lock I updated here is at the top level. The only one which got update. 
Lock file inside fuzzer is a bit confuse for me too. But It is not touched during the PR.

PR #150, opened 2023-02-01
sgx-verify: add function extract_tcb_info_from_raw_dcap_quote

OverOrion:
This function is used in the worker for extracting `FMSPC` from a raw DCAP quote.

### DIFF
--- a/teerex/sgx-verify/src/lib.rs
+++ b/teerex/sgx-verify/src/lib.rs
@@ -492,6 +492,27 @@ pub fn verify_certificate_chain<'a>(
 	Ok(leaf_cert)
 }
 
+pub fn extract_tcb_info_from_raw_dcap_quote(
+	dcap_quote_raw: &[u8],
+) -> Result<([u8; 6], TcbVersionStatus), &'static str> {
+	let mut dcap_quote_clone = dcap_quote_raw;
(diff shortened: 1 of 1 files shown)


teerex/sgx-verify/src/lib.rs
```diff
@@ -492,6 +492,27 @@ pub fn verify_certificate_chain<'a>(
 	Ok(leaf_cert)
 }
 
+pub fn extract_tcb_info_from_raw_dcap_quote(
+	dcap_quote_raw: &[u8],
+) -> Result<([u8; 6], TcbVersionStatus), &'static str> {
```
clangenb, line 497: I think we have defined a type alias for this.

clangenb (commented):
I only have one remark, otherwise this looks great! Verify readable, and the error messages are clear!

teerex/sgx-verify/src/lib.rs
OverOrion: You are right, I forgot about that when I was doing a rebase, than you!

clangenb (approved):
Thanks for the fix! Looks good to me know!

clangenb (approved):
Looks good to me!

PR #151, opened 2023-02-07
Add `publish_hash` extrinsic

### LINKED ISSUE #145: Publish (merkle-root) hash of proof of execution on the parachain
We do trust a TEE does things correctly, but the tee does need to prove that it did execute something at all. For indirect requests, we already do this; we publish the hash of the executed request on the chain. However, we don't do this yet for direct requests. This could be done with a very simple extrinsic:

```rust
pub fn publish_hash(origin, hahs: H256) {
    let mrenclave = ensure_registered_enclave(origin)?;

    // The enclave publishes a busines logic specific hash, which might be:
    // * merkle-root of some executed operations
    // * hash of a direct request that was sent aka `hash_of(trusted_call)`
    // * Other examples ??
    Self::deposit_event(Event::PublishedHash{enclave: mrenclave, hash: hash});
}
```

As we don't want to bloat our chain state, we only emit an event. All the indexers will still register and store the event, and you can always query archive nodes for events at a specific block height. The only thing that would not be possible is to request the hash via XCM, which I believe that we don't need it. In this case, the TEE would just send the result or whatever to the other parachain directly anyhow.

clangenb:
Closes #145

### DIFF
--- a/teerex/src/benchmarking.rs
+++ b/teerex/src/benchmarking.rs
@@ -62,7 +62,7 @@ benchmarks! {
 	// Benchmark `register_enclave` with the worst possible conditions:
 	// * remote attestation is valid
 	// * enclave already exists
-	where_clause {  where T::AccountId: From<[u8; 32]> }
+	where_clause {  where T::AccountId: From<[u8; 32]>, T::Hash: From<[u8; 32]> }
 	register_enclave {
 		ensure_not_skipping_ra_check();
 		timestamp::Pallet::<T>::set_timestamp(TEST4_SETUP.timestamp.checked_into().unwrap());
--- a/teerex/src/lib.rs
+++ b/teerex/src/lib.rs
@@ -473,7 +473,9 @@ impl<T: Config> Pallet<T> {
 		Ok(().into())
 	}
 
-	fn get_enclave(sender: &T::AccountId) -> Result<Enclave<T::AccountId, Vec<u8>>, Error<T>> {
+	pub(crate) fn get_enclave(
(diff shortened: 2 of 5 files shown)


teerex/src/lib.rs
```diff
@@ -367,6 +377,36 @@ pub mod pallet {
 			<TcbInfo<T>>::insert(fmspc, on_chain_info);
 			Ok(().into())
 		}
+
+		/// Public a hash as a result of an arbitrary enclave operation.
```
OverOrion, line 385:
```suggestion
		/// Publish a hash as a result of an arbitrary enclave operation.
```

teerex/src/benchmarking.rs
```diff
@@ -127,8 +127,9 @@ benchmarks! {
 
 	}: _(RawOrigin::Signed(account), [1u8; 32].into(), topics::<T>(), get_data(l))
 	verify {
-		// Event comparison in an actual node is way too cumbersome as event does not
-		// implement `PartialEq`. We test this in the regular tests anyhow.
+		// Event comparison in an actual node is way too cumbersome as the `RuntimeEvent`
+		// does not implement `PartialEq`. So we only verify that the event is emitted here,
+		// and we do more thorough checks are in the normal cargo tests.
```
OverOrion:
```suggestion
		// and we do more thorough checks in normal cargo tests.

```

OverOrion (approved):
LGTM, thanks @clangenb, very ~nice~ _clean_ job :100:

Issue #125, opened 2022-12-20
Benchmarking for DCAP attestation


Niederb:
For the implementation of DCAP I just copied the weights of the IAS implementation but we need to properly benchmark this.

PR #159, opened 2023-03-17
polkadot v0.9.38

clangenb:
Based on @BillyWooo's  work in #125, but fixing the xcm stuff.

Notable commit: https://github.com/integritee-network/pallets/pull/159/commits/78a0f30adb81b1e09927c9c2714f8527fe79742a the linker can't handle a wasmtime version mismatch between subxt (used in tests) and substrate, there is a duplicate function def error then. Luckily, we can disable the wasm stuff in subxt.

### DIFF
--- a/xcm-transactor/Cargo.toml
+++ b/xcm-transactor/Cargo.toml
@@ -25,10 +25,10 @@ sp-runtime = { default-features = false, git = "https://github.com/paritytech/su
 sp-std = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.38" }
 
 # xcm/polkadot
-xcm = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.38", default-features = false }
+xcm = { default-features = false, git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.38" }
 
 # cumulus
-cumulus-primitives-core = { git = "https://github.com/paritytech/cumulus", branch = "polkadot-v0.9.38", default-features = false }
+cumulus-primitives-core = { default-features = false, git = "https://github.com/paritytech/cumulus", branch = "polkadot-v0.9.38" }
 
 [dev-dependencies]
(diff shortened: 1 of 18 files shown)


xcm-transactor/Cargo.toml
```diff
@@ -17,27 +17,28 @@ scale-info = { version = "2.0.1", default-features = false, features = ["derive"
...
+externalities = { package = "sp-externalities", git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.38" }
+frame-benchmarking = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.38" }
 hex-literal = "0.3.2"
-pallet-balances = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.37" }
+pallet-balances = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.38" }
 test-utils = { path = "../test-utils" }
-sp-keyring = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.37" }
-subxt = { git = "https://github.com/paritytech/subxt.git", branch = "master" }
+sp-keyring = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.38" }
+# remove wasm stuff by disabling default features.
```
OverOrion, line 40: Could you please extend the comment with something that explains why? 
Like "or else there will be a wasmtime mismatch" or something like that.

OverOrion (approved):
LGTM, nice catch! :rocket:

xcm-transactor/Cargo.toml
clangenb: haha, I could link it here, but I thought, even without the version mismatch, we don't want this entire tree of dependencies if we don't need them. Do  you think I should still add it?

clangenb: Nevermind, I added it. Thanks for being picky!

PR #168, opened 2023-05-12
change license for teerex to MS-RSL

brenzi:
if this license actually prohibits forking our repo, we should extract teerex into its own repo. not sure TBH. but we can just go with this and change it if someone complains that forking is hereby prohibited

The basic goal is to be transparent, but protect integritee network from copycats on its basic value proposition (or DIY RA on other parachains)


clangenb (approved):
I don't know enough to actually judge the licence, but the way the new licence is added looks good to me.

PR #175, opened 2023-05-26
Polkadot update v0.9.42

OverOrion:
See the following substrate PRs for details:
- https://github.com/paritytech/substrate/pull/13535
- https://github.com/paritytech/substrate/pull/13475
- https://github.com/paritytech/substrate/pull/13798

For the last one I opted for suppressing the warning, but will open an issue for benchmarking it.

Closes https://github.com/integritee-network/pallets/issues/172

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,6 +69,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "android-tzdata"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
+
+[[package]]
 name = "android_system_properties"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -142,7 +148,7 @@ checksum = "b9ccdd8f2a161be9bd5c023df56f1b2a0bd1d83872ae53b71a84a12c9bf6e842"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.17",
+ "syn 2.0.18",
 ]
 
 [[package]]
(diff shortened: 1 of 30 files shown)


OverOrion:
Did the required followup changes, but now there is an internal substrate error:
``Self::mutate_account``` will fail with `NoProviders`.

The error originates from `inc_consumers`, where the `providers` field shoul not be 0.

Currently investigating why this happened.

Cargo.lock
```diff
@@ -511,24 +524,25 @@ version = "0.11.1"
...
  "polkadot-parachain",
  "polkadot-primitives",
+ "scale-info",
  "sp-api",
  "sp-runtime",
- "sp-std 5.0.0",
+ "sp-std 5.0.0 (git+https://github.com/paritytech/substrate?branch=polkadot-v0.9.42)",
  "sp-trie",
- "xcm",
+ "xcm 0.9.42",
```
clangenb, line 545: This should be investigated, we seem to have divergent substrate versions. Hint: Always check that the added and deleted lines are somewhat net-zero.

parentchain/src/lib.rs
```diff
@@ -9,7 +9,7 @@ pub mod pallet {
 	use frame_system::pallet_prelude::*;
 
 	#[pallet::pallet]
-	#[pallet::generate_store(pub(super) trait Store)]
+
```
clangenb, line 12: Pls. remove the empty line.

clangenb (commented):
I could imagine that the error comes from divergent substrate versions.

Cargo.lock
OverOrion: Thanks Chris for the tip, I believe I found the culprit:
`xcm` related things have `release-vX.Y.Z` branch names instead of `polkadot-vX.Y.Z`, which is why they were left untouched, resulting in duplicates.

clangenb: haha, happened to me too. :)

coax1d (commented):
All of this looks good to me. All the fun polkadot update stuff.

clangenb (approved):
Nice, looks good to me! Now we just have to fix the test.

PR #180, opened 2023-06-07
Add events for all dispatchables

### LINKED ISSUE #174: Events for all dispatchables
some dispatchables do not dispatch any Events. This should change:

register_tcb_info
* should dispatch Event `TcbInfoRegistered(fmspc)`

register_quoting_enclave
* should dispatch `QuotingEnclaveRegistered`. not yet sure about additional data needed

### LINKED ISSUE #179: enrich AddedEnclave Event with attestation method and TCB status
https://github.com/integritee-network/pallets/blob/e124aebb2d3d05a9a65f209f8e6304c6790f15d5/teerex/src/lib.rs#L77

What would be interesting to know in this event:
* was it DCAP or IAS registered?
* what was the TCB_status? `OK`, `SWHardeningNeeded` .....

coax1d:
closes #174
closes #179 

Some dispatchables are missing events..

### DIFF
--- a/teerex/src/lib.rs
+++ b/teerex/src/lib.rs
@@ -26,6 +26,7 @@ use frame_support::{
 use frame_system::{self, ensure_signed};
 use sgx_verify::{
 	deserialize_enclave_identity, deserialize_tcb_info, extract_certs, verify_certificate_chain,
+	SgxStatus,
 };
 use sp_core::H256;
 use sp_runtime::{traits::SaturatedConversion, Saturating};
@@ -83,6 +84,8 @@ pub mod pallet {
 		AddedEnclave {
 			registered_by: T::AccountId,
 			worker_url: Vec<u8>,
+			tcb_status: SgxStatus,
+			attestation_type: Attestation,
 		},
 		RemovedEnclave(T::AccountId),
 		Forwarded(ShardIdentifier),
@@ -172,13 +175,16 @@ pub mod pallet {
 			log::info!("teerex: parameter length ok");
 
 			#[cfg(not(feature = "skip-ias-check"))]
-			let enclave = Self::verify_report(&sender, ra_report).map(|report| {
-				Enclave::new(
-					sender.clone(),
-					report.mr_enclave,
-					report.timestamp,
-					worker_url.clone(),
-					report.build_mode,
+			let (enclave, report) = Self::verify_report(&sender, ra_report).map(|report| {
+				(
+					Enclave::new(
+						sender.clone(),
+						report.mr_enclave,
+						report.timestamp,
+						worker_url.clone(),
+						report.build_mode,
+					),
+					report,
 				)
 			})?;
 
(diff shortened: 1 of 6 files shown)


teerex/src/lib.rs
```diff
@@ -92,6 +92,12 @@ pub mod pallet {
 			hash: H256,
 			data: Vec<u8>,
 		},
+		TcbInfoRegistered {
+			fmspc: [u8; 6],
```
clangenb, line 96: We have a type for this FMSPC.

clangenb (approved):
Beautiful, simple and easy

teerex/src/lib.rs
coax1d: Great catch I missed that alias!

coax1d:
@clangenb I added an event for the parentchain pallet is this ok?

clangenb (approved):
Nice, locks great to me!

teerex/src/lib.rs
```diff
@@ -359,9 +365,13 @@ pub mod pallet {
...
 			let _sender = ensure_signed(origin)?;
-			let quoting_enclave =
-				Self::verify_quoting_enclave(enclave_identity, signature, certificate_chain)?;
+			let quoting_enclave = Self::verify_quoting_enclave(
+				enclave_identity.clone(),
+				signature,
+				certificate_chain,
+			)?;
 			<QuotingEnclaveRegistry<T>>::put(quoting_enclave);
+			Self::deposit_event(Event::QuotingEnclaveRegistered { enclave_identity });
```
clangenb, line 374: Hmm, sorry. I think I have approved to fast. I have been pondering about this, and I believe we should actually put the whole quoting enclave struct here instead of the encoded vec.

clangenb (changes requested):

teerex/src/lib.rs
```diff
@@ -92,6 +92,12 @@ pub mod pallet {
 			hash: H256,
 			data: Vec<u8>,
 		},
+		TcbInfoRegistered {
+			fmspc: Fmspc,
+		},
+		QuotingEnclaveRegistered {
+			enclave_identity: Vec<u8>,
+		},
```
clangenb, line 100: Conclusively I think this should be as follows:
```rust
TcbInfoRegistered {
	fmspc: Fmspc,
        tcb_info: TcbInfoOnChain
},
QuotingEnclaveRegistered {
	enclave_identity: QuotingEnclave,
},
```

Is there any downside of putting the whole data @brenzi? I don't think so.

teerex/src/lib.rs
coax1d: Will do

coax1d:
> > @clangenb I added an event for the parentchain pallet is this ok?
> 
> Yes very good!
> 
> I wonder if we also want to tackle this while we are at it: #179

I can add #179 into this one as well

coax1d:
Think i got the report handled in #179 lmk

brenzi (changes requested):

brenzi (approved):

clangenb (approved):
LGTM, thanks @coax1d for the fixes!

PR #181, opened 2023-06-22
Teeerex: Be much more verbose during the process.

OverOrion:


primitives/teerex/src/lib.rs
```diff
@@ -143,10 +143,16 @@ impl TcbVersionStatus {
 
 	pub fn verify_examinee(&self, examinee: &TcbVersionStatus) -> bool {
 		for (v, r) in self.cpusvn.iter().zip(examinee.cpusvn.iter()) {
+			log::debug!("teerex: verify_examinee: v={:#?},r={:#?}", v, r);
```
clangenb, line 146: Thanks a lot for all these extra logs. However, I am not fond of `teerex:`, it is essentially redundant information as the logger does already contain the rust module path. But we could add a `target: TEEREX` as an argument, where teerex should be a const ;)

teerex/sgx-verify/src/collateral.rs
```diff
@@ -230,6 +230,9 @@ impl TcbInfo {
 	}
 
 	pub fn is_valid(&self, timestamp_millis: i64) -> bool {
+		log::info!("teerex: called into runtime call register_tcb_info(), inside Self::is_valid.");
+		log::info!("teerex: called into runtime call register_tcb_info(), inside Self::is_valid, self is: {:#?}" , &self);
+		log::info!("teerex: called into runtime call register_tcb_info(), inside Self::is_valid, timestamp_millis: {:#?}" , &timestamp_millis);
```
clangenb, line 235: Together with the above suggestion of setting a target instead of just having the `teerex` in the string I suggest:
```suggestion
		log::info!(target: TEEREX, "Called into runtime call register_tcb_info(), inside Self::is_valid.");
		log::info!(target: TEEREX, "self is: {:#?}" , &self);
		log::info!(target: TEEREX, "timestamp_millis: {:#?}" , &timestamp_millis);
```

What do you think?

clangenb (commented):
Ah sorry, I just saw that you did not actually as for a review yet, so I stopped. :)

primitives/teerex/src/lib.rs
OverOrion: I did not know about this, handy feature! Rewrote all of them to use it, thanks for the early review! :D :100:

primitives/teerex/src/lib.rs
```diff
@@ -22,6 +22,8 @@ use scale_info::TypeInfo;
 use sp_core::H256;
 use sp_std::prelude::*;
 
+pub const TEEREX: &str = "TEEREX";
```
clangenb, line 25: I don't actually know if there is a specific reason, but in other codebases this seems to be lowercase.

teerex/sgx-verify/src/lib.rs
```diff
@@ -478,9 +479,27 @@ pub fn deserialize_tcb_info(
 	signature: &[u8],
 	certificate: &webpki::EndEntityCert,
 ) -> Result<TcbInfo, &'static str> {
+	log::info!(
+		target: TEEREX,
+		"Called into runtime call register_tcb_info(), inside Self::deserialize_tcb_info."
```
clangenb, line 484: I am not sure if `Called into runtime call ...` is right because this assumes some context, which is not always true. It should be evident from the sequence in the logs, that we entered from the `register_tcb_info`.

teerex/src/lib.rs
```diff
@@ -405,7 +406,7 @@ pub mod pallet {
 			signature: Vec<u8>,
 			certificate_chain: Vec<u8>,
 		) -> DispatchResultWithPostInfo {
-			log::info!("teerex: called into runtime call register_quoting_enclave()");
+			log::info!(target: TEEREX, "Called into runtime call register_quoting_enclave()");
```
clangenb, line 409: Don't you think most of these logs are trace?

clangenb (commented):
I think I would downgrade some logs to `trace!`, and also strip some of the info from some of the logs. I have mentioned two example cases, where I think this is true. :)

teerex/src/lib.rs
OverOrion: Absolutely, when I dropped the patches for the TCB Info V2 compatibility it seems I had dropped commits  I should have kept, sorry about this! :crying_cat_face:

teerex/sgx-verify/src/lib.rs
OverOrion: Yep, removed them!

teerex/src/lib.rs
```diff
@@ -638,19 +641,29 @@ impl<T: Config> Pallet<T> {
...
+		log::trace!(target: TEEREX, "TCB Info verification done, result is: {:#?}", &res);
+		// TODO reenable check
+		//ensure!(res, "tcb_info is outdated");
+
+		log::debug!(target: TEEREX, "DCAP quote ensured. tcbinfo: {:?}", &tcb_info);
 
 		let enclave_signer = T::AccountId::decode(&mut &report.pubkey[..])
 			.map_err(|_| <Error<T>>::EnclaveSignerDecodeError)?;
-		ensure!(sender == &enclave_signer, <Error<T>>::SenderIsNotAttestedEnclave);
+		// ensure!(sender == &enclave_signer, <Error<T>>::SenderIsNotAttestedEnclave);
```
clangenb, line 657: So we removed this because of the aforemention attesteer usecase?

clangenb (approved):
Thanks for the fixes! Looks good to me now!

teerex/src/lib.rs
OverOrion: That was a leftover as well, in retrospect I should have been more granular and created two branches :)

clangenb:
Sorry, for merging the other one first. I remember that Alain brought in a `log::info!("teerex: ***"), maybe you can fix this while resolving the conflicts. :)

brenzi:
can we await the merge of #191 and #193 plz?

brenzi (approved):

Issue #183, opened 2023-07-05
[teerex&sidechain] introduce ShardStatus


brenzi:
new storage in teerex
`ShardStatus: map ShardIdentifier -> Vec<AccountId, EnclaveFingerprint, BlockNumber>`

use cases (for sidechains and OCWs):
* show which instances are active
* show which instances have upgraded their fingerprint

* [x] sidechain pallet needs to update this storage upon new block confirmations

Issue #152, opened 2023-02-07
Unregister silent workers should not be called upon every timestamp


clangenb:
It is highly inefficient like this. And in general, I think we need to revise the procedure in order to be scalable.

brenzi:
to be replaced with lazy garbage collection

PR #191, opened 2023-07-07
teerex refactor 2

### LINKED ISSUE #186: [teerex] unify dispatchables for enclave registration
storage
image

image

image

dispatchables
image

brenzi:
closes #186 #183 #152 

breaking changes:
* ShardStatus will replace WorkerForShard logic
* remove skip-ias
* introduce MultiEnclave

non-breaking changes:
* no automatic garbage collection OnTimestamp anymore. lazy dispatchable introduced

### DIFF
--- a/primitives/teerex/Cargo.toml
+++ b/primitives/teerex/Cargo.toml
@@ -10,6 +10,7 @@ edition = "2021"
 [dependencies]
 codec = { version = "3.0.0", default-features = false, features = ["derive"], package = "parity-scale-codec" }
 common-primitives = { path = "../common", default-features = false }
+derive_more = "0.99.16"
 scale-info = { version = "2.0.1", default-features = false, features = ["derive"] }
 serde = { version = "1.0.13", default-features = false }
 
--- a/primitives/teerex/Cargo.toml
+++ b/primitives/teerex/Cargo.toml
@@ -16,6 +16,7 @@ serde = { version = "1.0.13", default-features = false }
 # substrate dependencies
 sp-core = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.42" }
 sp-io = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.42" }
+sp-runtime = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.42" }
 sp-std = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.42" }
 
 [dev-dependencies]
--- a/primitives/teerex/Cargo.toml
+++ b/primitives/teerex/Cargo.toml
@@ -32,5 +32,6 @@ std = [
     # substrate
     "sp-core/std",
     "sp-io/std",
+    "sp-runtime/std",
     "sp-std/std",
 ]
--- a/primitives/teerex/src/lib.rs
+++ b/primitives/teerex/src/lib.rs
@@ -19,7 +19,8 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 use codec::{Decode, Encode};
 use scale_info::TypeInfo;
-use sp_core::H256;
+use sp_core::{bounded_vec::BoundedVec, ConstU32, H256};
+use sp_runtime::MultiSigner;
 use sp_std::prelude::*;
 
 #[derive(Encode, Decode, Copy, Clone, PartialEq, Eq, sp_core::RuntimeDebug, TypeInfo)]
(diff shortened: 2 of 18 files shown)


brenzi:
I just realize that refactoring the sidechain pallet will likely require to solve #183 because we currently only let the first ever registered validateer finalize and the move to a storage map no longer let's us derive an order

brenzi:
breaking change: `ShardStatus` will replace `WorkerForShard` logic

primitives/teerex/Cargo.toml
```diff
@@ -16,6 +16,7 @@ serde = { version = "1.0.13", default-features = false }
 # substrate dependencies
 sp-core = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.42" }
 sp-io = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.42" }
+sp-runtime = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.42" }
```
clangenb, line 19: std enablement missing

primitives/teerex/src/lib.rs
```diff
@@ -92,6 +93,94 @@ pub enum SgxStatus {
...
+pub enum AnySigner {
+	Opaque(OpaqueSigner),
+	Known(MultiSigner),
+}
+
+impl From<MultiSigner> for AnySigner {
+	fn from(signer: MultiSigner) -> Self {
+		AnySigner::Known(signer)
+	}
+}
```
clangenb, line 109: We could remove these manual implementations by importing `derive_more` and then derive `From`

teerex/src/lib.rs
```diff
@@ -162,70 +169,143 @@ pub mod pallet {
...
+							T::AccountId::decode(&mut report.report_data.lower32().as_ref()),
+						<Error<T>>::SenderIsNotAttestedEnclave
+					);
+
+					// TODO: activate state checks as soon as we've fixed our setup #83
+					// ensure!((report.status == SgxStatus::Ok) | (report.status == SgxStatus::ConfigurationNeeded),
+					//     "RA status is insufficient");
+					// log::info!("teerex: status is acceptable");
+
+					Self::ensure_timestamp_within_24_hours(report.timestamp)?;
```
clangenb, line 213: Intuitively, it makes more sense if this is directly after verifying the report.

teerex/src/lib.rs
```diff
@@ -162,70 +169,143 @@ pub mod pallet {
...
+			ensure!(
+				worker_url.clone().unwrap_or_default().len() <= MAX_URL_LEN,
+				<Error<T>>::EnclaveUrlTooLong
+			);
 			log::info!("teerex: parameter length ok");
 
-			#[cfg(not(feature = "skip-ias-check"))]
-			let enclave = Self::verify_report(&sender, ra_report)?.with_url(worker_url.clone());
+			let enclave = match attestation_method {
+				SgxAttestationMethod::Ias => {
```
clangenb, line 188: This whole function will be much more readable if we have something like:

```rust
let enclave = match attestation_method {
   SgxAttestationMethod::Ias => verify_ias_attestation(...)?,
   SgxAttestationMethod::Dcap { proxied } => verify_dcap_quote(...)?,
   SgxAttestationMethod::Skipped { proxied } => handle_skipped_attestation(...)?
}


```

teerex/src/lib.rs
```diff
@@ -162,70 +169,143 @@ pub mod pallet {
...
+			log::info!("teerex: called into runtime call unregister_sovereign_enclave()");
 			let sender = ensure_signed(origin)?;
-
-			Self::remove_enclave(&sender)?;
-			Self::deposit_event(Event::RemovedEnclave(sender));
+			let enclave = Self::sovereign_enclaves(&enclave_signer)
+				.ok_or(<Error<T>>::EnclaveIsNotRegistered)?;
+			let now = <timestamp::Module<T>>::get();
+			let oldest_acceptable_attestation_time =
+				now.saturating_sub(T::MaxSilenceTime::get()).saturated_into::<u64>();
```
clangenb, line 304: I think we should rename this to `T::AttestationTimeout` or something like this.

teerex/src/lib.rs
```diff
@@ -474,7 +500,7 @@ pub mod pallet {
 		/// The worker url is too long.
 		EnclaveUrlTooLong,
 		/// The Remote Attestation report is too long.
-		RaReportTooLong,
+		RaProofTooLong,
```
clangenb, line 503: The doc should be adjusted as well

teerex/src/lib.rs
```diff
@@ -717,11 +597,45 @@ impl<T: Config> Pallet<T> {
...
+
+	pub fn most_recent_shard_update(
+		shard: &ShardIdentifier,
+	) -> Option<ShardSignerStatus<T::AccountId, T::BlockNumber>> {
+		if let Some(mut statuses) = <ShardStatus<T>>::get(shard) {
+			statuses.sort_by_key(|a| a.last_activity);
+			statuses.last().cloned()
+		} else {
+			None
+		}
```
clangenb, line 638:
```suggestion
		<ShardStatus<T>>::get(shard).map(|statuses| {
			statuses.sort_by_key(|a| a.last_activity);
			statuses.last().cloned()
		})
```

teerex/src/lib.rs
```diff
@@ -717,11 +597,45 @@ impl<T: Config> Pallet<T> {
...
+		let signer_statuses = if let Some(mut status_vec) = <ShardStatus<T>>::get(shard) {
+			if let Some(index) = status_vec.iter().position(|i| i.signer == *enclave_signer) {
+				status_vec[index] = fresh_status;
+			} else {
+				status_vec.push(fresh_status)
+			}
+			status_vec
+		} else {
+			vec![fresh_status]
+		};
```
clangenb, line 625:
```suggestion
		let signer_statuses = get_or_init(...);
		
// more efficient, and maybe even more readable. :)
pub fn get_or_init(...) {
  <ShardStatus<T>>::get(shard)
  .map(|status_vec| {
      status_vec.iter().position(|i| i.signer == *enclave_signer)
        .map_or_else(|| status_vec.push(free_status), |i| status_vec[i] = fresh_status);
      
      status_vec
  }).unwrap_or_else(|| vec![fresh_status])
}
```

teerex/src/lib.rs
```diff
@@ -717,11 +597,45 @@ impl<T: Config> Pallet<T> {
...
+	pub fn poke_shard(
+		shard: ShardIdentifier,
+		enclave_signer: &T::AccountId,
+	) -> Result<ShardSignerStatuses<T>, DispatchErrorWithPostInfo> {
+		let enclave = Self::sovereign_enclaves(enclave_signer.clone())
+			.ok_or(<Error<T>>::EnclaveIsNotRegistered)?;
+
+		let current_block_number = <frame_system::Pallet<T>>::block_number();
+
+		let fresh_status = ShardSignerStatus {
```
clangenb, line 610: `new_status` seems like a more common terminology to me.

teerex/src/lib.rs
```diff
@@ -717,11 +597,45 @@ impl<T: Config> Pallet<T> {
 			Err(<Error<T>>::RemoteAttestationTooOld.into())
 		}
 	}
-}
 
-impl<T: Config> OnTimestampSet<T::Moment> for Pallet<T> {
-	fn on_timestamp_set(moment: T::Moment) {
-		Self::unregister_silent_workers(moment)
+	pub fn poke_shard(
```
clangenb, line 601: From the outside it is a bit hard to guess, what is `poke`, I think the more commonly used term `touch` reflects better, what this function represents.

clangenb (commented):
Reviewed everything except for the tests. Looks good so far!

teerex/src/lib.rs
brenzi: Thanks for the intermediate review. I'll iterate and guess its better to wait with the next phase of the review

teerex/src/lib.rs
brenzi: I think readability is debatable. I don't like 3 line dispatchables that do nothing else than calling a fn within the same file....and then dispatch an event. if it's only a few lines I prefer reading them in-place.
Testability would be a stronger argument IMO, but as all unit tests go to the dispatchable level directly, I saw no reason to modularize

teerex/src/lib.rs
brenzi: max silence time also renews upon activity. not only attestation. We could use MaxPassiveTime?

primitives/teerex/src/lib.rs
brenzi: much wow! love it!

teerex/src/lib.rs
brenzi: borrow checker not happy with this. but I adapted pragmatically ;-)

teerex/src/lib.rs
brenzi: not much better IMO

clangenb: I also thought that this is the design, but when I looked at the code, I could only find that it is renewed upon a new attestation. But in this case, I guess I was wrong, and then it is ok as it is

teerex/src/lib.rs
clangenb: Hmm, I disagree here to be honest, when going down, it needs a lot of focus to even remember, whether you are still in a match branch or not, but I won't argue.

teerex/src/lib.rs
```diff
@@ -162,70 +169,147 @@ pub mod pallet {
...
-			log::info!("teerex: called into runtime call register_ias_enclave()");
+			log::info!("teerex: called into runtime call register_sgx_enclave()");
 			let sender = ensure_signed(origin)?;
-			ensure!(ra_report.len() <= MAX_RA_REPORT_LEN, <Error<T>>::RaReportTooLong);
-			ensure!(worker_url.len() <= MAX_URL_LEN, <Error<T>>::EnclaveUrlTooLong);
+			ensure!(proof.len() <= SGX_RA_PROOF_MAX_LEN, <Error<T>>::RaProofTooLong);
+			ensure!(
+				worker_url.clone().unwrap_or_default().len() <= MAX_URL_LEN,
+				<Error<T>>::EnclaveUrlTooLong
+			);
```
clangenb, line 184: Unnecessary clone can be removed.

Maybe `worker_url.as_ref()` needed in the `if let` statement.
```suggestion
if let Some(url) = worker_url {
	ensure!(url.len() <= MAX_URL_LEN, <Error<T>>::EnclaveUrlTooLong);
}
```

clangenb (approved):
Alright, looks good to me except for one detail, but I shall approve already. :)

PR #193, opened 2023-07-08
teerex refactor 3

brenzi:


primitives/teerex/src/lib.rs
```diff
@@ -103,14 +103,42 @@ pub enum AnySigner {
...
 }
+impl Default for AnySigner {
+	fn default() -> Self {
+		AnySigner::Opaque(OpaqueSigner::default())
+	}
+}
+
+impl From<[u8; 32]> for AnySigner {
+	fn from(pubkey: [u8; 32]) -> Self {
+		// zero padding is necessary because the chain storage does that anyway for bounded vec
```
brenzi, line 114: this concerns me. if we don't zero-pad, the tests fail. but why are BoundedVecs bloated by the onchain state backend?

PR #204, opened 2023-07-15
introduce convenient handling of attestation skip and debug mode security flags

brenzi:
* clean up remainders of the skip-ias-check compiler flag
* introduce bool storage value for allowing skipping attestion
* introduce a root origin dispatchable to set the security flags


teerex/src/benchmarking.rs
```diff
@@ -87,7 +78,6 @@ benchmarks! {
 	// Benchmark `register_tcb_info` with the worst possible conditions:
 	// * tcb registration succeeds
 	register_tcb_info {
-		ensure_not_skipping_ra_check();
```
clangenb, line 90: Need to make sure that the benchmarks runtime does not skip this otherwise, we will get weights, which are orders of magnitude weaker. To make sure that the benchmark is correct, we should probably overwrite the fields here, regardless of what is put in the genesis config: https://github.com/integritee-network/pallets/issues/219

clangenb (commented):
One minor detail that we could fix, otherwise it looks good!

teerex/src/benchmarking.rs
brenzi: I see your point

PR #215, opened 2023-08-07
add proxied field to ias attestation_method too for attesteer rpc

### LINKED ISSUE #216: add proxied field to Ias attestation_method too
because the attesteer rpc is actually capable of doing just that

brenzi:
closes #216 
this is a state breaking change. not sure when to merge

without this PR, IAS attestation through attesteer rpc will fail because the enclave signer isn't the sender of the registration extrinsic


clangenb (commented):
LGTM, shouldn't we merge this now, before we upgrade rococo, so we only have one set of breaking changes?

brenzi:
I thought about this, but there's a whole set of upgrades in consequence, through multiple repositories. If this PR stays the only breaking change, I'd rather save that time because this feature is unrequested

brenzi:
IAS is discontinued anyway. we can drop this

PR #220, opened 2023-08-08
state migrations to teerex v1

brenzi:
simply purge teerex registry

### DIFF
--- a/teerex/src/migrations.rs
+++ b/teerex/src/migrations.rs
@@ -1,0 +1,150 @@
+use super::*;
+
+use frame_support::{pallet_prelude::*, storage_alias, traits::OnRuntimeUpgrade};
+
+/// The log target.
+const TARGET: &str = "teerex::migration::v1";
+
+mod v0 {
+	use super::*;
+
+	#[derive(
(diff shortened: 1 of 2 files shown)


teerex/src/migrations.rs
```diff
@@ -0,0 +1,150 @@
...
+use frame_support::{pallet_prelude::*, storage_alias, traits::OnRuntimeUpgrade};
+
+/// The log target.
+const TARGET: &str = "teerex::migration::v1";
+
+mod v0 {
+	use super::*;
+
+	#[derive(
+		Encode, Decode, Default, Copy, Clone, PartialEq, Eq, sp_core::RuntimeDebug, TypeInfo,
```
clangenb, line 12: I don't think it really matters, but I guess we don't need any of these derives for the old types? However, deriving from comments of parity engineers I read, I think that the rust compiler doesn't even include unneeded implementations of types any more - so it doesn't hurt at all.

clangenb (approved):
LGTM

teerex/src/migrations.rs
brenzi: true.

PR #221, opened 2023-08-31
unit test longer alibaba quote

brenzi:


brenzi:
closing in favor of #225 which includes this commit

PR #223, opened 2023-09-12
Explicit errors for sgx_verify

brenzi:
neither substrate_api_client nor js/apps properly enrich nested errors (although we derive PalletError and TypeInfo), but at least we propagate an error code which we can be matched to the enum manually

So, a signature error on sgx_verify now looks like this:
`[2023-09-12T14:54:00Z ERROR integritee_cli::base_cli::commands::register_tcb_info] register_tcb_info extrinsic failed Dispatch(Module(ModuleError { pallet: "Teerex", error: "SgxVerifyError", description: ["An error originating in the sgx_verify crate"], error_data: ModuleErrorData { pallet_index: 50, error: [15, 34, 0, 0] } })) `

or:
image

34 (or 2nd byte in raw error `0x22`) is the index of the `sgx_verify::Error::RsaSignatureInvalid`

### DIFF
--- a/teerex/sgx-verify/src/lib.rs
+++ b/teerex/sgx-verify/src/lib.rs
@@ -522,7 +522,7 @@ pub fn deserialize_tcb_info(
 /// Extract a list of certificates from a byte vec. The certificates must be separated by
 /// `-----BEGIN CERTIFICATE-----` and `-----END CERTIFICATE-----` markers
 pub fn extract_certs(cert_chain: &[u8]) -> Vec<Vec<u8>> {
-	// The certificates should be valid UTF-8 but if not we continue. The certificate verification
+	// The certificates should be valid UTF-8 but if not we skip the invalid cert. The certificate verification
 	// will fail at a later point.
 	let certs_concat = String::from_utf8_lossy(cert_chain);
 	let certs_concat = certs_concat.replace('\n', "");
--- a/teerex/sgx-verify/src/lib.rs
+++ b/teerex/sgx-verify/src/lib.rs
@@ -64,6 +64,54 @@ pub mod test_data;
 mod tests;
 mod utils;
 
+#[derive(Debug, Encode, Decode, Copy, Clone, TypeInfo, PartialEq)]
+pub enum Error {
+	/// invalid RSA signature
+	RsaSignatureInvalid,
+	/// TCB info could not be deserialized
+	TcbInfoInvalid,
+	///
+	TimestampInvalid,
+	TimestampMissing,
+	///
+	QuoteStatusMissing,
+
+	QuoteBodyInvalid,
+	QuoteBodyMissing,
+	QuoteBodyDecodingError,
+	KeyLengthInvalid,
+	PublicKeyInvalid,
+	DerEncodingError,
+	SgxReportParsingError,
+	DcapQuoteVersionMismatch,
+	DcapKeyTypeMismatch,
+	DcapQuoteTooLong,
+	DcapQuoteDecodingError,
+	PckCertFormatMismatch,
+	QeRejectedEnclave,
+	CertificateChainTooShort,
+	QeReportHashMismatch,
+	IsvEnclaveReportSignatureInvalid,
+	FmspcLengthMismatch,
+	FmspcDecodingError,
+	FmspcOidMissing,
+	CpuSvnLengthMismatch,
+	CpuSvnOidMissing,
+	CpuSvnDecodingError,
+	PceSvnOidMissing,
+	PceSvnLengthMismatch,
+	PceSvnDecodingError,
+	EnclaveIdentityDecodingError,
+	EnclaveIdentitySignatureInvalid,
+	LeafCertificateParsingError,
+	CertificateChainInvalid,
+	NetscapeDecodingError,
+	NetscapeDerError,
+	IntelExtensionCertificateDecodingError,
+	IntelExtensionAmbiguity,
+	CaVerificationFailed,
+}
+
 #[derive(Debug, Encode, Decode, Copy, Clone, TypeInfo)]
 #[repr(C)]
 pub struct SGXAttributes {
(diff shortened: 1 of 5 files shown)


teerex/sgx-verify/src/lib.rs
```diff
@@ -731,25 +775,23 @@ pub fn verify_signature(
...
-			#[cfg(test)]
-			println!("IAS signature is valid");
+			log::trace!(target: TEEREX, "RSA signature is valid");
 			Ok(())
 		},
 		Err(_e) => {
-			#[cfg(test)]
-			println!("RSA Signature ERROR: {}", _e);
-			Err("bad signature")
+			log::info!(target: TEEREX, "RSA Signature ERROR: {}", _e);
```
clangenb, line 785: remove the underscore

teerex/sgx-verify/src/lib.rs
```diff
@@ -494,32 +538,32 @@ pub fn verify_certificate_chain<'a>(
...
 	let leaf_cert: webpki::EndEntityCert =
-		webpki::EndEntityCert::from(leaf_cert).map_err(|_| "Failed to parse leaf certificate")?;
+		webpki::EndEntityCert::from(leaf_cert).map_err(|_| Error::LeafCertificateParsingError)?;
 	log::trace!(target: TEEREX, "inside Self::verify_certificate_chain, leaf cert parsed.");
 	let time = webpki::Time::from_seconds_since_unix_epoch(verification_time / 1000);
 	let sig_algs = &[&webpki::ECDSA_P256_SHA256];
 	leaf_cert
 		.verify_is_valid_tls_server_cert(sig_algs, &DCAP_SERVER_ROOTS, intermediate_certs, time)
-		.map_err(|_| "Invalid certificate chain")?;
+		.map_err(|_| Error::CertificateChainInvalid)?;
```
clangenb, line 550: dito

teerex/sgx-verify/src/lib.rs
```diff
@@ -533,24 +577,24 @@ pub fn verify_dcap_quote(
...
 		quote.quote_signature_data.qe_certification_data.certification_data_type == 5,
-		"Only support for PEM formatted PCK Cert Chain"
+		Error::PckCertFormatMismatch //Only support for PEM formatted PCK Cert Chain
 	);
-	ensure!(quote.quote_signature_data.qe_report.verify(qe), "Enclave rejected by quoting enclave");
+	ensure!(quote.quote_signature_data.qe_report.verify(qe), Error::QeRejectedEnclave); //"Enclave rejected by quoting enclave"
 
 	let certs = extract_certs(&quote.quote_signature_data.qe_certification_data.certification_data);
-	ensure!(certs.len() >= 2, "Certificate chain must have at least two certificates");
+	ensure!(certs.len() >= 2, Error::CertificateChainTooShort); //"Certificate chain must have at least two certificates"
```
clangenb, line 597: Here to `..isTooShort`

teerex/sgx-verify/src/lib.rs
```diff
@@ -533,24 +577,24 @@ pub fn verify_dcap_quote(
...
-	ensure!(quote.header.attestation_key_type == 2, "Only support for ECDSA-256");
+	ensure!(quote.header.version == 3, Error::DcapQuoteVersionMismatch);
+	ensure!(quote.header.attestation_key_type == 2, Error::DcapKeyTypeMismatch);
 	ensure!(
 		quote.quote_signature_data.qe_certification_data.certification_data_type == 5,
-		"Only support for PEM formatted PCK Cert Chain"
+		Error::PckCertFormatMismatch //Only support for PEM formatted PCK Cert Chain
 	);
-	ensure!(quote.quote_signature_data.qe_report.verify(qe), "Enclave rejected by quoting enclave");
+	ensure!(quote.quote_signature_data.qe_report.verify(qe), Error::QeRejectedEnclave); //"Enclave rejected by quoting enclave"
```
clangenb, line 594: To disambiguate we maybe want to have: `QeHasRejectedEnclave`

clangenb (commented):
Looks good to me, some minor nitpicks.

There is only one idea that I have actually seen in substrate I believe, we could transform the `sgx_verify::Error` with a conversion trait:

```rust
impl From<sgx_verify::Error> for Error {
    from(e: sgx_verify::Error) -> Self {
       // put all the variants we care about explicitly
       match e {
          // The pallet's error needs to replicate all error variants we care about
          CaVerificationFailed => Self::CaVerificationFailed,
          ...
          /// just use a default statement for the erros we don't really care about explicitly.
          other_error => Self::OtherSgxVerifyError(other_error),
       }
    }
}
```

This means redundant code, but it will be supported by all tooling, so maybe it is worth it.

teerex/sgx-verify/src/lib.rs
```diff
@@ -420,22 +463,22 @@ pub struct CertDer<'a>(&'a [u8]);
 
 /// Encode two 32-byte values in DER format
 /// This is meant for 256 bit ECC signatures or public keys
-pub fn encode_as_der(data: &[u8]) -> Result<Vec<u8>, &'static str> {
+pub fn encode_as_der(data: &[u8]) -> Result<Vec<u8>, Error> {
 	if data.len() != 64 {
-		return Result::Err("Key must be 64 bytes long")
+		return Result::Err(Error::KeyLengthInvalid)
```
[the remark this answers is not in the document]
brenzi: I did it this way exactly to sort  by topic. but I don't mind adding an "Is"

clangenb:
Fixing the cargo about installation properly in a separate PR ;) https://github.com/integritee-network/pallets/pull/224

teerex/src/lib.rs
```diff
@@ -105,27 +105,45 @@ pub mod pallet {
...
+
+	impl<T> From<sgx_verify::Error> for Error<T> {
+		fn from(e: sgx_verify::Error) -> Self {
+			use sgx_verify::Error as Theirs;
+			match e {
+				Theirs::CaVerificationFailed => Self::SgxCaVerificationFailed,
+				Theirs::RsaSignatureIsInvalid => Self::SgxRsaSignatureIsInvalid,
+				_ => Self::OtherSgxVerifyError(e),
+			}
+		}
```
brenzi, line 146: tried to impl your suggestion @clangenb  but the compiler can't find this implementation of From:
```
the trait `From<sgx_verify::Error>` is not implemented for `sp_runtime::DispatchError`
    |
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
    = help: the following other types implement trait `From<T>`:
              <sp_runtime::DispatchError as From<&'static str>>
              <sp_runtime::DispatchError as From<ArithmeticError>>
              <sp_runtime::DispatchError as From<TokenError>>
              <sp_runtime::DispatchError as From<TransactionalError>>
              <sp_runtime::DispatchError as From<frame_system::Error<T>>>
              <sp_runtime::DispatchError as From<hidden_include::error::BadOrigin>>
              <sp_runtime::DispatchError as From<hidden_include::error::LookupError>>
              <sp_runtime::DispatchError as From<pallet::Error<T>>>
              <sp_runtime::DispatchError as From<pallet_balances::Error<T, I>>>
    = note: required for `sgx_verify::Error` to implement `Into<sp_runtime::DispatchError>`
    = note: required for `DispatchErrorWithPostInfo<PostDispatchInfo>` to implement `From<sgx_verify::Error>`
    = note: required for `Result<([u8; 6], teerex_primitives::SgxTcbInfoOnChain), DispatchErrorWithPostInfo<PostDispatchInfo>>` to implement `FromResidual<Result<Infallible, sgx_verify::Error>>`

```

brenzi: expanded the macros but couldn't find the cause

brenzi: this shouldn'r be necessary, but it works if I'm very explicit:
`.map_err(|e| Error::<T>::from(e))?`
not as nice as a simple `?`, but I can live with this

clangenb: I believe that the problem is that we are doing a two layered conversion and I think the compiler can't handle that: `sgx_verify::Error -> Error<T> -> sp_runtime::DispatchError`

I am quite positive that this doesn't have to do anything with the macro magic, but the rust compiler simply can't do that.

clangenb: The only way to fix this would be to introduce a wrapper function, but I don't think it is worth it:

```rust
fn verify_ias_report(..) -> Error<T> {
     sgx_verify::sgx_verify_ias_report(...)?:
}

// somewhere in the pallet:
Self::verify_ias_report()?;
```

brenzi: the readability is better as-is IMO

clangenb (approved):
Nice, I very much appreciate this improvement.

clangenb:
The PR description doesn't match the implementation now.

PR #228, opened 2023-11-17
Update `webpki` and `ring-xous` dependency

Kailai-Wang:
Pretty much a copy-up of https://github.com/litentry/litentry-parachain/pull/2229, with additional adjustments of error types and `scale-info` crate update.

If we need to use the pallets in no-std, we need to override the `ring` crate, like [here](https://github.com/litentry/litentry-parachain/pull/2229/files#diff-7e5f583905509761b08c323cca20f5f283e5544a7ab5e3eb18c598af0f281de0):

```
[patch.crates-io]
ring = { git = "https://github.com/betrusted-io/ring-xous", branch = "0.16.20-cleanup" }
```


clangenb:
@Kailai-Wang master has the fix for the subxt test now, feel free to update your branch.

Kailai-Wang:
> @Kailai-Wang master has the fix for the subxt test now, feel free to update your branch.

Thanks, I updated it. It should be ready for review now.

teerex/Cargo.toml
```diff
@@ -12,8 +12,9 @@ edition = "2021"
 codec = { version = "3.0.0", default-features = false, features = ["derive"], package = "parity-scale-codec" }
 hex = { default-features = false, version = "0.4.3", features = ["alloc"] }
 log = { version = "0.4.14", default-features = false }
-scale-info = { version = "2.0.1", default-features = false, features = ["derive"] }
+scale-info = { version = "2.10.0", default-features = false, features = ["derive"] }
 serde = { version = "1.0", features = ["derive"], optional = true }
+webpki = { git = "https://github.com/rustls/webpki", version = "=0.102.0-alpha.3", rev = "da923ed", package = "rustls-webpki", default-features = false, features = ["alloc", "ring"] }
```
clangenb, line 17: We did not have a code change in this pallet, why do we have a new dependency here?

teerex/sgx-verify/src/lib.rs
```diff
@@ -378,21 +378,17 @@ pub struct SgxVerifiedReport {
 	pub build_mode: SgxBuildMode,
 }
 
-type SignatureAlgorithms = &'static [&'static webpki::SignatureAlgorithm];
+type SignatureAlgorithms = &'static [&'static dyn webpki::types::SignatureVerificationAlgorithm];
```
clangenb, line 381: Can you explain to me why this is dyn?

teerex/sgx-verify/src/lib.rs
```diff
@@ -378,21 +378,17 @@ pub struct SgxVerifiedReport {
...
 static SUPPORTED_SIG_ALGS: SignatureAlgorithms = &[
-	//&webpki::ECDSA_P256_SHA256,
-	//&webpki::ECDSA_P256_SHA384,
-	//&webpki::ECDSA_P384_SHA256,
-	//&webpki::ECDSA_P384_SHA384,
```
clangenb, line 386: This is something that I didn't notice before. Why do we comment the ECDSA signature algorithms, when we are verify signatures below with it. @brenzi I guess this is you who did it?

clangenb (commented):
Nice, looks good, except for one fundamental question, I have only one comment. Otherwise, I like the change, it makes the code more readable, and I also like that we can get rid of our fork.

You only need to fix the toml-fmt, and I would like to ask you to mention the patch in the readme. Thanks a lot for the contribution!

teerex/sgx-verify/src/lib.rs
brenzi: Don't remember. I guess our patch was incomplete at the time

teerex/Cargo.toml
Kailai-Wang: We do have [changes](https://github.com/integritee-network/pallets/pull/228/files#diff-b00e78f3cf2794df255a5ca3a7682559534cddf0c0986aff24d680e051abc422) in `teerex/src/lib.rs`

teerex/sgx-verify/src/lib.rs
Kailai-Wang: Because `SignatureVerificationAlgorithm` is a trait? And `webpki::ring::RSA_PKCS1_2048_8192_SHA256` (and the alike) is a trait object by itself: `dyn SignatureVerificationAlgorithm`

Kailai-Wang:
> Nice, looks good, except for one fundamental question, I have only one comment. Otherwise, I like the change, it makes the code more readable, and I also like that we can get rid of our fork.
> 
> You only need to fix the toml-fmt, and I would like to ask you to mention the patch in the readme. Thanks a lot for the contribution!

Thanks, I fixed that!
If you are happy with the text/position in README, please feel free to adjust it

teerex/Cargo.toml
clangenb: Ah yes. The git diff was not really clear, all good!

teerex/sgx-verify/src/lib.rs
clangenb: But it seems that this has changed in this case, can't we use an enum/struct like before? I will accept and merge the PR anyhow, I am just curious.

clangenb (approved):
Nice, looks good to me!

teerex/sgx-verify/src/lib.rs
Kailai-Wang: but the `verify_for_usage` looks for something like this:
```
    pub fn verify_for_usage(
        &self,
        supported_sig_algs: &[&dyn SignatureVerificationAlgorithm],
```

Not sure if that's what you look for tho...

clangenb: Ah, I see, then I guess this was indeed an api change. Sorry for having been bothering you, while just being too lazy to dig in the dependency crate. Thanks a lot!

Kailai-Wang: All good!

brenzi:
this causes trouble when I try to integrate it, even with the suggested patch:
image

brenzi:
copying the Cargo.lock file from litentry fixed the mentioned issue

PR #236, opened 2023-12-22
accept any new sidechain block from authorized enclave for shard

### LINKED ISSUE #235: how to recover from ReceivedUnexpectedSidechainBlock
the sidechain finalization quick hack is very brittle:

https://github.com/integritee-network/pallets/blob/94aa66fbe63d0d5bc848b9cf8f7a71fa34ef8ce8/sidechain/src/lib.rs#L117-L120

If for any reason that single extrinsic that should supply the expected sidechain block gets lost (i.e. denied because of an enclave update causing WrongFingerprintForShard) we can never recover

brenzi:
closes #235 

the previous logic where validateers had to announce their next block to be expected was too brittle and had no effective security benefit. All we need is:
* only authorized enclaves may finalize
* block numbers must increase monotonically
* avoid forks

avoiding forks is still the naive solution: just finalize blocks from the first validateer appearing in the ShardStatus, which usually is the primary validateer who created the shard

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -873,7 +873,7 @@ dependencies = [
 [[package]]
 name = "frame-benchmarking"
 version = "4.0.0-dev"
-source = "git+https://github.com/paritytech/substrate.git?branch=polkadot-v0.9.42#ff24c60ac7d9f87727ecdd0ded9a80c56e4f4b65"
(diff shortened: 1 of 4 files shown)


clangenb (approved):
Hmm, I see the point of avoiding forks by choosing which validateer is authorized to have its blocks finalized, and we can already say that having a round robing scheme for this authorization, we kind of ensure quick finality, if we don't have many validateers.

Might have been overengineered before.

PR #247, opened 2024-02-01
cleanup workspace

brenzi:
before migrating to polkadot 1.3.0 and crates.io, move all dependency definitions to workspace toml to remove redundancy and simplify future upgrades

functionally, this PR is a noop


clangenb (approved):
Wonderful. Thanks for making this a separate PR and removing this awful dependency renamings.

PR #248, opened 2024-02-01
upgrade to polkadot v1.3.0

brenzi:
and depend on crates.io

that was lightweight :-)


clangenb (approved):
Nice, the cargo.lock looks good. It seems like no duplicates were introduced.

PR #255, opened 2024-02-23
implement SBliFF

### LINKED ISSUE #242: SBliFF v1: Semi-Blind Faith Finality V1
The most simple finality for Integritee TEE-validated sidechains only guarantees the following:
1. next finalized block is a lineal descendant of previous finalized block
2. block number is monotonically increasing
3. ShardConfig is respected
   * only authenticated TEE's within a potentially limited set are authorized to propose finality candidate blocks
   
from the chain's perspective, the validateers are trusted to only supply legit blocks. They are, however, not trusted to be up to date with either the chain state nor its peers

Therefore, there is no need to know intermediate block hashes to accept new blocks. We can relax (1) and just make sure the validateer knows all blocks between our last finalized block and his proposed candidate block

# BliFF specification

storage:
* last_finalized_block: (BlockNumber, BlockHash)

dispatchables:

```
confirm_imported_sidechain_block(
  shard,
  latest_finalized_ancestor: (BlockNumber, BlockHash),
  finalization_candidate: (BlockNumber, BlockHash),
)

ensure(authorized for shard)
if finalization_candidate.number <= last_finalized_block.number: fail
if latest_finalized_ancestor.number != last_finalized_block.number: fail
if latest_finalized_ancestor.hash != last_finalized_block.hash: fail

last_finalized_block = finalization_candidate
```

brenzi:
closes #242 

must bump tx version

### DIFF
--- a/sidechain/src/lib.rs
+++ b/sidechain/src/lib.rs
@@ -116,22 +116,20 @@ pub mod pallet {
 			}
 
 			if let Some(known_ancestor) = Self::latest_sidechain_block_confirmation(shard) {
-				if let Some(provided_ancestor) = latest_finalized_ancestor {
-					ensure!(
-						finalization_candidate.block_number > known_ancestor.block_number,
-						<Error<T>>::FinalizationCandidateIsOutdated
-					);
-					ensure!(
-						known_ancestor.block_number == provided_ancestor.block_number,
-						<Error<T>>::AncestorNumberMismatch
-					);
-					ensure!(
-						known_ancestor.block_header_hash == provided_ancestor.block_header_hash,
-						<Error<T>>::AncestorHashMismatch
-					);
-				} else {
-					return Err(Error::<T>::AncestorMissing.into())
-				}
+				let provided_ancestor =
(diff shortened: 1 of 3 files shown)


sidechain/src/lib.rs
```diff
@@ -100,25 +105,35 @@ pub mod pallet {
...
+						known_ancestor.block_number == provided_ancestor.block_number,
+						<Error<T>>::AncestorNumberMismatch
+					);
+					ensure!(
+						known_ancestor.block_header_hash == provided_ancestor.block_header_hash,
+						<Error<T>>::AncestorHashMismatch
+					);
+				} else {
+					return Err(Error::<T>::AncestorMissing.into())
+				}
```
clangenb, line 134: I think it is easier to read with
```rust
let provided ancestor = latest_finalized_ancestor.ok_or_else(|| return Err(Error::<T>::AncestorMissing.into())?;
ensure!(...);
ensure!(...);
```

clangenb (approved):
Fairly straight forward, looks good to me

PR #263, opened 2024-07-04
pallet teerdays

brenzi:
branded as TEERdays, but could be a sober "tokentime" just as well. The pallet is generic for all native tokens.

We're introducing a new non-transferrable token "TEERdays" which can be earned by bonding TEER
TEERdays shall serve as a governance token for L2 shards (i.e. Incognitee). 
Multiplying bond with time makes governance more resilient against deep pocket attackers who suddenly decide to allocate a lot of TEER to take over governance. The time factor acts like inertia to governance.

This pallet is only the minimum necessary to start collecting TEERdays

notes:
* numeric stability:
   * 100years in milliseconds (Moment) are 42bits
   * 10MTEER with 12 digits are 60bits
   * 100years of total supply locked still fits u128
   * therefore, it is safe to use the Balance type for TEERdays

This PR also changes the toolchain to stable, except rustfmt


teerdays/src/tests.rs
```diff
@@ -0,0 +1,289 @@
...
+		});
+		assert!(System::events().iter().any(|a| a.event == expected_event));
+
+		let teerdays = TeerDays::teerday_bonds(&alice)
+			.expect("TeerDays entry for bonded account should exist");
+		assert_eq!(teerdays.value, amount - unbond_amount);
+		assert_eq!(
+			teerdays.accumulated_tokentime,
+			tokentime_accumulated.saturating_mul(amount - unbond_amount) / amount
+		);
```
clangenb, line 196: Maybe some comment here would be nice

`// accumulated tokentime is reduced pro-rata`

teerdays/src/tests.rs
```diff
@@ -0,0 +1,289 @@
...
+
+		// can't unbond again
+		assert_noop!(
+			TeerDays::unbond(RuntimeOrigin::signed(alice.clone()), unbond_amount),
+			Error::<Test>::PendingUnlock
+		);
+		// withdrawing not yet possible. fails silently
```
clangenb, line 203: Why silently? It seems that an error is returned.

teerdays/src/tests.rs
```diff
@@ -0,0 +1,289 @@
...
+		assert_eq!(teerdays.accumulated_tokentime, amount * 10_000);
+		assert_eq!(teerdays.last_updated, now);
+
+		let account_info = System::account(&alice);
+		assert_eq!(account_info.data.frozen, amount + 1_000_000_000_000);
+	})
+}
+
+#[test]
+fn unbonding_and_delayed_withdraw_works() {
```
clangenb, line 161: For me would be more concise to say:

`withrawing_unbonded_after_unlock_period_works`

teerdays/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+		#[pallet::weight(< T as Config >::WeightInfo::bond())]
+		pub fn bond_extra(
+			origin: OriginFor<T>,
+			#[pallet::compact] value: BalanceOf<T>,
+		) -> DispatchResult {
+			let signer = ensure_signed(origin)?;
+			ensure!(value >= T::Currency::minimum_balance(), Error::<T>::InsufficientBond);
+			let bond = Self::do_update_teerdays(&signer)?;
+			let free_balance = T::Currency::free_balance(&signer);
+			// free confusingly includes the already bonded amount, so we need to subtract it
```
clangenb, line 202: Is this for real? Doesn't that hint that we might make a mistake during the bonding phase?

teerdays/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+		}
+
+		#[pallet::call_index(2)]
+		#[pallet::weight(< T as Config >::WeightInfo::unbond())]
+		pub fn unbond(
+			origin: OriginFor<T>,
+			#[pallet::compact] value: BalanceOf<T>,
+		) -> DispatchResult {
+			let signer = ensure_signed(origin)?;
+			ensure!(Self::pending_unlock(&signer).is_none(), Error::<T>::PendingUnlock);
```
clangenb, line 223: This might be todo for the future. I think it is not very convenient that we can't unbond more if we are already unbonding, but I do agree that this introduces quite some more complexity.

teerdays/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+				.saturating_mul(new_bonded_amount);
+
+			let new_bond = TeerDayBondOf::<T> {
+				value: new_bonded_amount,
+				last_updated: now,
+				accumulated_tokentime: new_tokentime,
+			};
+
+			if new_bond.value < T::Currency::minimum_balance() {
+				TeerDayBonds::<T>::remove(&signer);
```
clangenb, line 245: Should be documented that bonding is subject to an ED.

teerdays/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+		let bond = bond.update(now);
+		TeerDayBonds::<T>::insert(account, bond);
+		Self::deposit_event(Event::<T>::TokenTimeUpdated { account: account.clone(), bond });
+		Ok(bond)
+	}
+
+	fn try_withdraw_unbonded(
+		account: &T::AccountId,
+	) -> Result<BalanceOf<T>, sp_runtime::DispatchError> {
+		if let Some((due, amount)) = Self::pending_unlock(account) {
```
clangenb, line 298: We can remove the nesting here with :

```rust
let (due,amount) = Self::pending_unlock(account).ok_or_else(|| Error::<T>::NotUnlocking)?;
```

clangenb (commented):
Looks good to me, no unclarities.

But I do understand correctly that, so far, the Teerdays are not relevant elsewhere?

teerdays/src/lib.rs
brenzi: seems to be the same in the staking pallet. very confusing. But given the unit tests, it does what we expect. Will double check how js/apps displays this once we integrate it into the node

teerdays/src/lib.rs
brenzi: yeah. the first implementation shall be of minimal complexity - even if that comes with some UX downsides

brenzi:
thank you. You're right: TEERdays are consumed nowhere. The nomination pallet is left to do where TEERdays will be applied

primitives/teerdays/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+use parity_scale_codec::{Decode, Encode};
+use scale_info::TypeInfo;
+use sp_runtime::{traits::AtLeast32BitUnsigned, Saturating};
+
+#[derive(Encode, Decode, Copy, Clone, PartialEq, Eq, Default, sp_core::RuntimeDebug, TypeInfo)]
+pub struct TeerDayBond<Balance, Moment> {
+	pub value: Balance,
+	pub last_updated: Moment,
+	// the unit here is actually balance * moments.
+	pub accumulated_tokentime: Balance,
```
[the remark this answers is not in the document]
brenzi: that would just add a third type parameter and look overly verbose to me because it's an associated type in the end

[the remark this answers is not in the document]
brenzi: unsure....will leave it for now

teerdays/src/weights.rs
```diff
@@ -0,0 +1,63 @@
...
+	fn unbond() -> Weight;
+	fn update_other() -> Weight;
+	fn withdraw_unbonded() -> Weight;
+}
+
+/// Weights for pallet_sidechain using the Integritee parachain node and recommended hardware.
+pub struct IntegriteeWeight<T>(PhantomData<T>);
+impl<T: frame_system::Config> WeightInfo for IntegriteeWeight<T> {
+	fn bond() -> Weight {
+		todo!()
```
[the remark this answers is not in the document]
brenzi: I suggest to first run benchmarks and then feedback meaningful default values here:
https://github.com/integritee-network/parachain/issues/298

teerdays/src/tests.rs
```diff
@@ -0,0 +1,289 @@
...
+		set_timestamp(now);
+		let alice = AccountKeyring::Alice.to_account_id();
+		let alice_free: BalanceOf<Test> = 5_000_000_000_000;
+		<Test as pallet::Config>::Currency::make_free_balance_be(&alice, alice_free);
+		let amount: BalanceOf<Test> = 10_000_000_000_000;
+		assert_ok!(TeerDays::bond(RuntimeOrigin::signed(alice.clone()), amount));
+
+		let expected_event = RuntimeEvent::TeerDays(TeerDaysEvent::Bonded {
+			account: alice.clone(),
+			amount: alice_free,
```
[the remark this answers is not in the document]
brenzi: I saw a similar `min()` logic in pallet_staking and went along. I don't think it hurts as the users' intention is clear

teerdays/src/tests.rs
brenzi: outdated comment, thx

teerdays/src/lib.rs
brenzi: added a todo here: https://github.com/integritee-network/parachain/issues/298

brenzi: you were right. there is an issue: https://github.com/integritee-network/pallets/pull/265

PR #264, opened 2024-07-19
fixing benchmarks

brenzi:
our benchmarks failed with mismatch in slot to timestamp.
this fixes it

### DIFF
--- a/teeracle/src/mock.rs
+++ b/teeracle/src/mock.rs
@@ -57,14 +57,6 @@ frame_support::construct_runtime!(
 	}
 );
 
-impl pallet_aura::Config for Test {
-	type AuthorityId = AuraId;
-	type DisabledValidators = ();
-	type MaxAuthorities = ConstU32<32>;
-	type AllowMultipleBlocksPerSlot = ConstBool<false>;
-	type SlotDuration = pallet_aura::MinimumPeriodTimesTwo<Test>;
-}
-
 parameter_types! {
 	pub const BlockHashCount: u32 = 250;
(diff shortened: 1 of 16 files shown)


brenzi:
FYI @clangenb

teerex/sgx-verify/src/collateral.rs
```diff
@@ -15,7 +15,6 @@
 
 */
 
-#![cfg_attr(not(feature = "std"), no_std)]
 pub extern crate alloc;
```
clangenb, line 18: This line can also be removed.

teeracle/src/mock.rs
```diff
@@ -117,7 +129,7 @@ pub type Moment = u64;
 
 impl pallet_timestamp::Config for Test {
 	type Moment = Moment;
-	type OnTimestampSet = ();
+	type OnTimestampSet = Aura;
```
clangenb, line 132: Hmm, I see. This whole PR looks like an overkill, but I guess we can remove the feature flag now in our runtimes at least 🥳

clangenb: Uff, I see it is gone now: https://github.com/integritee-network/integritee-node/blob/1a18046482c72620898300e04b29fa7a72e7a3ff/runtime/src/lib.rs#L289

Here we used to have a feature flag that set `OnTimestamp= ()` in case we were running the benchmark because Aura didn't like it when we messed with time. But I guess your implementation now, is better than the runtime hack we did before anyhow.

PR #265, opened 2024-07-20
[teerdays] explicitly add burned teerdays to event

brenzi:
for easier testing on testnet


brenzi:
this isn't actually an issue. the error was that the 12 decimals were missing in the amount.
In fact, frozen funds ARE part of free - but not of "transferrable"

PR #277, opened 2025-06-10
upgrade sdk to stable2503-6

brenzi:
bump ring to 0.17.7 (to be patched to ring-xous which has a 0.17.7 branch)

don't use rust 1.84.1 just yet (issues with new wasm target and getrandom, used by ring). only bump to 1.82.0


Cargo.toml
```diff
@@ -32,6 +32,7 @@ derive_more = "0.99.16"
 env_logger = "0.9.0"
 hex = { version = "0.4.3", default-features = false, features = ["alloc"] }
 hex-literal = "0.4.1"
+getrandom = { version = "0.2.10", default-features = false, features = ["js"] }
```
clangenb, line 35: Pretty sure that the feature JS should not be here, and has been introduced due to a wrong feature needed in std.

But maybe I am wrong.

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.81.0"
+channel = "1.82.0"
```
clangenb, line 2: Wrong PR description

clangenb (approved):
LGTM except for my tiny comments

Cargo.toml
brenzi: I introduced it because getrandom docs say the flag is needed for wasm32-unknown-unknown target

brenzi:
When building the parachain runtime, we get an impasse with ring-xous 0.17.7 because the minimum needed for other deps is 0.17.8 which hasn't been merged in ring-xous yet

teerex/sgx-verify/Cargo.toml
```diff
@@ -15,6 +15,7 @@ workspace = true
 base64 = { workspace = true }
 chrono = { workspace = true }
 der = { workspace = true }
+getrandom = { workspace = true }
```
clangenb, line 18: I think this is the problem. I see that you added that line in order to fix the wasm build issue in the parachain. However, I think that the problem is that an std-leak of sorts resulted in the compiler error requesting the `js` feature.

PR #292, opened 2025-08-26
add local fee to XcmFeePrarams

### LINKED ISSUE #290: [porteer] charge fees that the sovereign account has to pay on all hops
See discussion: https://github.com/integritee-network/parachain/pull/325#discussion_r2290706088

Reasons for changing fees:
* unbalances DEXes on AHs (better to balance DEXes)
* global change of exchange rates among relevant tokens TEER, KSM, DOT
* runtime upgrades changing weights
* upstream configuration changes (bridge relayer fee)

In any case, it is very likely that multihop fees need to be changeds at the same time, so we can extend `XcmFeeConfig` with a field `local`

brenzi:
closes #290

### DIFF
--- a/porteer/src/lib.rs
+++ b/porteer/src/lib.rs
@@ -112,8 +112,13 @@ pub mod pallet {
 	/// or
 	/// 2. AHP -> AHK -> IK
 	pub struct XcmFeeParams<Balance> {
+		/// to be paid by the `port_tokens` caller to cover all subsequent fees
+		pub local: Balance,
+		/// fees to be paid by sovereign account for source side Asset Hub execution involving swapping to KSM/DOT [TEER]
 		pub hop1: Balance,
+		/// fees to be paid by sovereign account for destination side Asset Hub execution involving swapping to DOT/KSM [KSM/DOT]
 		pub hop2: Balance,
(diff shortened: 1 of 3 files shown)


porteer/src/lib.rs
```diff
@@ -112,8 +112,13 @@ pub mod pallet {
...
 	/// 2. AHP -> AHK -> IK
 	pub struct XcmFeeParams<Balance> {
+		/// to be paid by the `port_tokens` caller to cover all subsequent fees
+		pub local: Balance,
+		/// fees to be paid by sovereign account for source side Asset Hub execution involving swapping to KSM/DOT [TEER]
 		pub hop1: Balance,
+		/// fees to be paid by sovereign account for destination side Asset Hub execution involving swapping to DOT/KSM [KSM/DOT]
 		pub hop2: Balance,
+		/// fees to be paid by sovereign account for destination side Integritee execution involving swapping to TEER [DOT/KSM]
 		pub hop3: Balance,
```
clangenb, line 122: I just noticed that the struct fields are not compact encoded. Shall we update that, or we do not care?

brenzi: fixed it below

PR #294, opened 2025-08-26
Ab/porteer optimize compact

brenzi:
the nonce and the balances arguments can and should be compact encoded

XcmFeeParams has compact fields but can't derive compact because only single-field structs could derive that. Maybe it works anyway? If not, no big harm done


clangenb:
> the nonce and the balances arguments can and should be compact encoded
> 
> XcmFeeParams has compact fields but can't derive compact because only single-field structs could derive that. Maybe it works anyway? If not, no big harm done

Yes, structs can never be compact as a whole, they are just the concatenation of their encoded fields. So you could argue that a struct that only has compact fields is implicitly compact too, but it will never satisfy the trait bound (and there is no need to).

PR #296, opened 2025-09-01
polkadot stable2506 1 dev

clangenb:


clangenb:
Found the reason for cancelling. It has to do with the output of zepter for some reason.

Running taplo fmt after zepter fixes the issue.

